PR #169, opened 2015-06-13
Fix  environment variables.

piyo7:
Fix environment variables in .env.example to accord with \test\integration\sendgrid.test.js.
```js
var API_USER    = process.env.SENDGRID_USERNAME || 'some_sendgrid_username';
var API_KEY     = process.env.SENDGRID_PASSWORD || 'some_sendgrid_password';
```


eddiezane:
Thanks!

piyo7:
Thanks for merging!

PR #179, opened 2015-07-26
Require just used functions & use natives

luanmuniz:
Hi,

I made this because after a few requests using this module, lodash was causing some memory allocation problems, the full file is very big and it was causing me problems.

I don't quite understand why, but make just these changes, requiring just the used functions and using a few native functions, my application decrease the memory allocation quite a lot.

### DIFF
--- a/lib/email.js
+++ b/lib/email.js
@@ -2,7 +2,8 @@
 
 var FileHandler     = require('./file_handler');
 var smtpapi_lib     = require('smtpapi');
-var _               = require('lodash');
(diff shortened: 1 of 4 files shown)


thinkingserious:
@luanmuniz,

Thanks for the pull request!

This is no longer a dependency in the new library: https://github.com/sendgrid/sendgrid-nodejs/tree/v3beta. It will replace this one in a few weeks.

Thanks for your support!

With Best Regards,

Elmer

Issue #162, opened 2015-05-01
.addCc("email@domian.com) not sending mail 


jtkinser:
When using .addCc("email@domain.com), the email is never actually sent to the CC address. However it is listed in the email CC field on the email that is sent to the .addTo() recipient.

eddiezane:
There is a breaking change that needs to be implemented in this library that will default to using our Web API parameters as opposed to currently using the SMTP API Header for `to` email addresses. The SMTP API Header does not support `cc` or `bcc` addresses.

A fix for the time being is to explicitly use the Web API parameters:

```javascript
email.to = ['foo@example.com', 'bar@example.com'];
email.addCc('baz@example.com'); // same as email.cc = 'baz@example.com'
```

Appologies for the confusion.

juanpgaviria:
Hi,

I have the same issue with cc and bcc, i try with email.cc and email.addCc and no email its sent to the cc addres..

eddiezane:
Hi @juanpgaviria,

I just confirmed that it does indeed work with these methods.

```javascript
var sg = require('sendgrid')(process.env.SENDGRID_APIKEY);

var email = new sg.Email();

email.to = 'foo@example.com';
email.subject = 'Testing CC + BCC';
email.from = 'me@example.com';
email.text = 'Testing';
email.addCc('bar@example.com');
email.bcc = 'baz@example.com';

sg.send(email, function(err, json) {
  if (err) console.error(err);
  console.log(json);
});
```

Please test with the latest version of the library and share a code sample if it still doesn't work.

juanpgaviria:
@eddiezane 

it works.. i was adding the bcc as array.. i think that the issue.. after 3 jack daniel's anything can happend.

thanks ;).

eddiezane:
Haha no worries! I realized that we were missing `addBcc` and `setBccs` so I just released a new version - 1.8.0 with them.

jcksncllwy:
Hi there! I believe I may have found a related bug. If email.addTo("foo@example.com") is called before email.addCc('bar@example.com') then the cc email is not sent.

```
var sg = require('sendgrid')(process.env.SENDGRID_APIKEY);

var email = new sg.Email();

email.addTo('foo@example.com');
email.subject = 'Testing CC + BCC';
email.from = 'me@example.com';
email.text = 'Testing';
email.addCc('bar@example.com');
email.bcc = 'baz@example.com';

sg.send(email, function(err, json) {
[...]

PR #187, opened 2015-09-09
Change to param to use Web API instead of X-SMTPAPI header

tushdante:
Hey @eddiezane . Here's the changes I've made. Would love it if you could give it a quick look through for me.

Thanks!


tushdante:
Also, closes #162

eddiezane:
The first failing test, `should not have multiple TOs if as an array but also set on smtpapi via addTo`, can probably be removed. No idea what that was even testing for.

The second one, `should have a to address if there is no to but there is an smtpapi to` was needed because the request would fail if there was no `to` param provided - even when using smtpapi. We would set it to the `from` in that case. Can probably be reworked and reworded for the new way.

tushdante:
So basically for the second one, `expect(JSON.parse(format['x-smtpapi']).to).to.not.be.empty;` needs to be removed cause we're not setting any `to` fields in the `smtpapi` headers anyway.

tushdante:
@thinkingserious so it looks like the build is failing on node v0.8 due to issues with either travis or node itself. i'm not sure how to procceed

thinkingserious:
Found the answer here: https://github.com/npm/npm/issues/9668 :/

tushdante:
Sweet! Thanks @thinkingserious Gotta finish up a few more tests and it should be ready for a code review

elbuo8:
This PR looked great guys. What I didn't see was, if the user only uses `smtpapi` recipients, no fake `Web API` recipient is set. That would cause it to break and not send the email. (This might have been fixed, but most libs have workarounds for that). Make sure you test sending just using `smtpapi` on dev env.

thinkingserious:
Thanks for the extra set of eyes @elbuo8. Testing now :)

thinkingserious:
Successfully tested using:

```
var sg = require('./lib/sendgrid')('<my_api_key>');
var email = new sg.Email();
email.addSmtpapiTo('elmer.thomas@gmail.com');
email.subject = 'Testing via SMTPAPI only';
email.from = 'dx@sendgrid.com';
email.text = 'Testing';
sg.send(email, function(err, json) {
  if (err) console.error(err);
  console.log(json);
});
```

elbuo8:
I missed this https://github.com/sendgrid/sendgrid-nodejs/blob/master/lib/email.js#L225

That's the hack that makes it work. Awesomeness :+1:

PR #209, opened 2016-01-28
Fix 'this' binding in initialization.

gpoesia:
Use _this instead of this to bind to _send on initialization.

I needed this change for the library to work here. Otherwise, in send(), this.options would be undefined because 'this' wasn't bound correctly. This caused the problem that no URI would be passed to request, which would complain with an error like "options.uri is a required option".

Does this work for everyone else? Because without this change I couldn't use the library at all, since all calls to send() would fail with that error.

I'm using node v0.10.25 and sendgrid 2.0.0. This node version is a bit old, but theoretically allowed by sendgrid's package.json, which only asks for a version >= 0.4.7.

Thanks.

### DIFF
--- a/lib/sendgrid.js
+++ b/lib/sendgrid.js
@@ -30,7 +30,7 @@ var Sendgrid = function(apiUserOrKey, apiKeyOrOptions, options) {
   }
 
   var _this         = this;
-  
+
   // do this to mantain similarity to other libs
   var uriParts = {};
   uriParts.protocol = this.options.protocol || "https";
@@ -51,7 +51,7 @@ var Sendgrid = function(apiUserOrKey, apiKeyOrOptions, options) {
       email = new Email(email);
     }
 
-    _send.bind(this)(email, callback);
+    _send.bind(_this)(email, callback);
   };
 
   var _send = function(email, callback) {
@@ -73,7 +73,7 @@ var Sendgrid = function(apiUserOrKey, apiKeyOrOptions, options) {
       var json;
 
       if(err) return callback(err, null);
-      
+
       try {
         json = JSON.parse(body);
       } catch (e) {


corymsmith:
@gpoesia I am seeing the exact same thing with Node 5.6 and had some people reported the same issue here: https://github.com/feathersjs/feathers-sendgrid/issues/3

corymsmith:
@charlottestjohn Any plans on merging this?

charlottestjohn:
@corymsmith - I'll look into it! The PR certainly looks straightforward enough :).

corymsmith:
That would be fantastic, thanks! I haven't tested this PR myself, just ran into the same issue with the options being undefined.

charlottestjohn:
@corymsmith - I asked the guys who own this repo - and they said it's on their plate, but not assigned to this sprint. It will probably be a bit before they can look at new pull requests, but if they can, they'll try and squeeze this one in! @thinkingserious is who you'll want to follow up with!

corymsmith:
Sounds great, thanks for looking into it @charlottestjohn

herkyl:
Any updates on this? Cannot use sendgrid without it.

thinkingserious:
Hello @gpoesia, @corymsmith, @herkyl,

This library will be replaced within a few weeks with this one: https://github.com/sendgrid/sendgrid-nodejs/tree/v3beta

I'm not sure if this issue will persist in the new library, as it's a complete rewrite. So I'll leave this open until further testing.

If you find that changes are still needed, I would need the pull request against the v3beta branch along with a signed CLA: https://github.com/sendgrid/sendgrid-nodejs/blob/v3beta/CONTRIBUTING.md#cla

Thank you for your patience and support!

With Best Regards,

Elmer

PR #230, opened 2016-04-29
switch to mime-types package

### LINKED ISSUE #207: Webpack support
I'm new to Webpack (apologies in advance for anything incorrect I say), and so not exactly sure how to proceed with this fix, but it seems is though the modules `fs`, `tls`, and `net` are unable to be resolved when bundling sendgrid with Webpack.

Has anyone else run into this problem?  If so, how did you resolve it?

**EDIT**:  Perhaps it is as easy as adding the following to the package.json?

```
  browser: {
    child_process: false
  }
```

chadsmith:
We're Webpack to bundle/minify dependencies for AWS Lambda, but were unable to to include Sendgrid due to the older version of `node-mime`.

This switches to `mime-types` + `mime-db` and should also resolve #207.

### DIFF
--- a/lib/file_handler.js
+++ b/lib/file_handler.js
@@ -1,7 +1,7 @@
 "use strict";
 
 var http = require('http');
-var mime = require('mime');
+var mime = require('mime-types');
 var _ = require('lodash');
 var url = require('url');
 var fs = require('fs');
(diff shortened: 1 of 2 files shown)


thinkingserious:
Hello @chadsmith,

This is no longer a dependency in the new library: https://github.com/sendgrid/sendgrid-nodejs/tree/v3beta. It will replace this library within a few weeks.

Thanks for your support!

With Best Regards,

Elmer

PR #251, opened 2016-06-27
Add TypeScript typings definition file for v3 API

SPARTAN563:
Hi guys,

Recently updated an application of ours to use your 3.0.0 library and noticed there weren't any up to date typings files in the wild. Seeing as the current approach with TypeScript is to embed these definition files in their respective NPM modules, I've taken the time to put together a compatible set of type definitions for your library.

In addition to adding the raw type definitions, I've also taken the liberty of adding a quick test that ensures the definition file is well formatted, it should help ensure that things aren't broken accidentally going forward.

As far as manual verification of the typings file goes, aside from running your unit test suites through the TypeScript compiler to confirm that they comply with the typings file (always a good way to ensure the types match what's expected), we're also running these typings on a production system with no issues as yet.

I'll ensure I sign the contributor license agreement and get that through to you guys, in the mean time if you wouldn't mind taking a look and letting me know if there's anything that stands out as being incorrect, or raising any issues you may have with this addition to the codebase.

## Things that may be issues
 - There's now a devDependency on the TypeScript compiler. While it isn't significantly large, I notice you've tried to keep the number of dependencies you have to a minimum (build times I assume being the reason). We can easily remove it should you wish, however that will remove the ability to test that the definitions are indeed valid.
 - There's something else to manage as the API changes. Unfortunately that's not something one can do much about (aside from switching to TypeScript for the library, not really an option). What I've done previously with projects on which we hold a dependency is offer to assist in updating the definitions whenever the API does change. In exchange I just ask that any of the project leads try and keep me informed (using a CC) whenever they are aware of any API changes.

Kind regards and all the best,
Benjamin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ node_js:
 - "4"
 before_install:
 - npm install -g npm@2
+script:
+- npm test
+- npm run test:typescript
 notifications:
   hipchat:
     rooms:
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,472 @@
+declare namespace SendGrid.Helpers.Mail {
+    export interface Helper {
+        Email: typeof Email;
+        Mail: typeof Mail;
+        Personalization: typeof Personalization;
+        Content: typeof Content;
+        Header: typeof Header;
+        Substitution: typeof Substitution;
+        CustomArgs: typeof CustomArgs;
+        Attachment: typeof Attachment;
+        Section: typeof Section;
+        Category: typeof Category;
+        Asm: typeof Asm;
+        Bcc: typeof Bcc;
+        BypassListManagement: typeof BypassListManagement;
+        Footer: typeof Footer;
+        SandBoxMode: typeof SandBoxMode;
+        SpamCheck: typeof SpamCheck;
+        MailSettings: typeof MailSettings;
+        ClickTracking: typeof ClickTracking;
+        OpenTracking: typeof OpenTracking;
+        SubscriptionTracking: typeof SubscriptionTracking;
+        Ganalytics: typeof Ganalytics;
+        TrackingSettings: typeof TrackingSettings;
+    }
+
+    export class Mail {
+        constructor();
+        constructor(from: Email, subject: string, to: Email, content: Content);
+
+        getFrom(): Email;
+        setFrom(email: Email): void;
+
+        addPersonalization(personalization: Personalization): void;
+        getPersonalizations(): Personalization[];
+
+        setSubject(subject: string): void;
+        getSubject(): string;
+
+        addContent(content: Content): void;
+        getContents(): Content[];
+
+        addAttachment(attachment: Attachment): void;
+        getAttachments(): Attachment[];
+
+        setTemplateId(templateId: string): void;
+        getTemplateId(): string;
+
+        addSection(section: Section): void;
+        getSections(): Section[];
+
+        addHeader(header: Header): void;
+        getHeaders(): Header[];
+
+        addCategory(category: Category): void;
+        getCategories(): Category[];
+
+        addCustomArg(customArg: CustomArgs): void;
+        getCustomArgs(): CustomArgs[];
+
+        setSendAt(sendAt: number): void;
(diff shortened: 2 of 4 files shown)


SPARTAN563:
Hey guys, just noticed that all the built in unit tests failed with a 401 error - suspecting that has something to do with your test suite environment variables as I haven't touched anything on that front - however if there's anything you'd like me to look at, please don't hessitate to ask.

SPARTAN563:
Hey guys, any word on whether this'll get included or whether I should instead be investigating other approaches to publishing it? Still having to hack our way around the lack of a usable type definition from any official source.

thinkingserious:
Hello @SPARTAN563,

It's still on our backlog and we plan to merge it. 

The only way this can move up our backlog queue is through +1's or comments on this thread.

Thanks for checkin in!

SPARTAN563:
I see, not a problem at all so long as there's a process in place - let me know if you need anything from me at any point and I'll do my best to help out.

CelsoSantos:
I too am writing an application in TypeScript and also need to use SendGrid. It's not that it's impossible to use without the Definition file, but it does become much harder has one has to constantly check on documentation instead of getting intellisense (in visual studio code) completions.

thinkingserious:
Thanks for the upvote @CelsoSantos, it helps move this issue up our queue.

thinkingserious:
Hi @SPARTAN563,

Could you please resolve the conflicts so that I may test and merge? Thanks!

SPARTAN563:
Hi @thinkingserious, should be up to date, rebased on `master`.

thinkingserious:
@SPARTAN563 please email us at dx@sendgrid.com with your T-shirt size and mailing address :)

frederickfogerty:
Hey @SPARTAN563 thanks for adding this to the lib, I LOVE it when modules have type definitions in their core. 🙌 

I'm having a problem when trying to use this module in my TypeScript project. I'm receiving this error when exporting a `SendGrid` instance: `Exported variable 'sendGrid' has or is using name 'SendGrid.SendGrid' from external module "[...]/node_modules/sendgrid/index" but cannot be named.`

This is a repro:

```
import { SendGrid } from 'sendgrid';
import * as SG from 'sendgrid'; // This is what we usually do to try squash the error, it doesn't work in this scenario :(

const sendGrid = SendGrid(process.env.SENDGRID_API_KEY);

export default sendGrid;
```

And since `SendGrid` is a namespace, I can't access it import it (?!), and thus can't explicitly type `sendGrid` with something like `const sendGrid: SendGrid.SendGrid = ...`.

🙏  🙌  Any help would be greatly appreciated, don't make us fall back to `any`! 😱

SPARTAN563:
Hi @frederickfogerty, just spun up an essentially empty project and wasn't able to reproduce the issue you're describing.

I've put down the example code you gave me in a quick repo which demonstrates this, if you could please fork it and make any changes necessary on your side to reproduce the issue then I'd be more than willing to look into it further.

Repo: https://github.com/SPARTAN563/sendgrid-ts-demo

Regards,
Benjamin

frederickfogerty:
Hey @SPARTAN563, thanks for taking a look at my issue, I really appreciate it.

I have submitted a PR to your repo which reproduces the issue: https://github.com/SPARTAN563/sendgrid-ts-demo/pull/1

Thanks!

thinkingserious:
@frederickfogerty,

Could the issue be that you are using `sendGrid` instead of `SendGrid`?

frederickfogerty:
Hey @thinkingserious, thanks for looking at my issue.

By using `SendGrid` instead of `sendGrid`, I'm assuming you mean for the variable name? In that case, there would be duplicate (i.e. conflicting) identifiers and wouldn't compile.

i.e. this wouldn't compile
```ts
import { SendGrid } from 'sendgrid';
import * as SG from 'sendgrid'; // This is what we usually do to try squash the error, it doesn't work in this scenario :(

const SendGrid = SendGrid(process.env.SENDGRID_API_KEY);
//    ^^^^^^^^ error here

export default SendGrid;
```

PR #299, opened 2016-09-19
Fill 'response' property in error instance

sitegui:
This makes the behavior when using callbacks consistent with the promise counterpart.
That is, in both cases, `error.response` is present

### DIFF
--- a/lib/sendgrid.js
+++ b/lib/sendgrid.js
@@ -111,6 +111,7 @@ SendGridInstance.prototype.API = function(request, callback) {
     }
     else {
       var error = new SendGridError('Response error');
+      error.response = response;
       callback(error, response);
     }
   });


sitegui:
Multiple fests failed with invalid http status response, I don't know why ;(

thinkingserious:
@sitegui,

Thanks again for another PR!

The link I sent you on the other pull request will apply here too. We'll just need it signed once.

To test locally, please follow steps 3-6 here: https://github.com/sendgrid/sendgrid-ruby/blob/master/CONTRIBUTING.md#initial-setup

They are not passing in Travis CI, because the tests only run for those in the SendGrid team. We are fixing that, and for now, after I get your signed CLA, I'll run the tests locally also as part of my pull review.

PR #300, opened 2016-09-20
fix: Correct TypeScript definitions

SPARTAN563:
See #251 and <https://github.com/SPARTAN563/sendgrid-ts-demo/pull/1> for more details on what wasn't working.

This ensures that the following code works correctly:

````typescript
import * as SendGrid from "sendgrid";

const sendGrid = SendGrid("MYAPIKEY");
``

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -456,6 +456,8 @@ declare namespace SendGrid {
     export interface SendGridConstructor {
         (apiKey: string, host?: string, globalHeaders?: { [header: string]: string; }): SendGrid;
         constructor(apiKey: string, host?: string, globalHeaders?: { [header: string]: string; }): SendGrid;
+
+        mail: SendGrid.Helpers.Mail.Helper;
     }
 
     export class SendGrid {
@@ -467,6 +469,5 @@ declare namespace SendGrid {
     }
 }
 
-declare const mail: SendGrid.Helpers.Mail.Helper;
 declare const sendGrid: SendGrid.SendGridConstructor;
-export {mail, sendGrid as SendGrid};
+export = sendGrid;

### CODE REVIEW

File: index.d.ts [hunk 1]
```diff
@@ -467,6 +469,5 @@ declare namespace SendGrid {
     }
 }
 
-declare const mail: SendGrid.Helpers.Mail.Helper;
 declare const sendGrid: SendGrid.SendGridConstructor;
-export {mail, sendGrid as SendGrid};
\ No newline at end of file
+export = sendGrid;
```
thinkingserious: I'm thinking this would be a breaking change. SendGrid is the expected case.

thinkingserious:
@SPARTAN563,

Thanks for the quick turnaround on the fix! I just had a few points of clarification in the review. We should be able to merge this one quickly.

index.d.ts
SPARTAN563: It is certainly a breaking change when it comes to the definition, however it corrects the definition to match the actual behaviour of the SendGrid JS library.

I've updated the interface above to include the rest of the library's exports as expected.

Javierenrique00:
Could you please add an example using TypeScript in NodeJS ?

ss108:
Hey,

Sorry to comment on a closed PR, but doesn't `emptyRequest` need to take an object as a parameter?

SPARTAN563:
Yeah, you're correct - sorry I missed that, let me quickly throw together a new PR

ss108:
I forked this and made a couple other changes to the definition: https://github.com/bloomtri/sendgrid-nodejs

The callback for the `API` method should be optional, and it needs to return a Promise. If those look good to you, then please incorporate them and I can delete that fork. 

Thanks for getting to this so promptly!

Issue #312, opened 2016-10-06
Send a Single Email to a Single Recipient


thinkingserious:
Acceptance Criteria:
- A email object that represents the response body payload
- A mail object that handles sending email objects, data validation and error handling

Reference:
- https://github.com/sendgrid/sendgrid-nodejs/tree/master/lib/helpers/mail
- https://github.com/sendgrid/sendgrid-nodejs/blob/master/CONTRIBUTING.md
- https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/index.html

Puigcerber:
Hi @thinkingserious!

Trying to find my fourth PR for this year #hacktoberfest I've seen [those exclusive pins](https://sendgrid.com/blog/happy-hacktoberfest/) and I would like to give it a go.

As I've never used Sendgrid before I'm a bit lost on where to start and understanding the issue. Is about extending the mail helper? Documenting how to send a single mail? I would love some clarifications.

Cheers, Pablo.

thinkingserious:
Hello Pablo,

First, I suggest you review the entire "Mail Helper Enhancement (v3 mail/send)" [project](https://github.com/sendgrid/sendgrid-nodejs/projects).

We want to refactor the mail helper so that it makes executing the use cases defined in this project as simple as possible.

For example now our hello world to send a single email looks like this:

```javascript
var helper = require('sendgrid').mail;
var from_email = new helper.Email('test@example.com');
var to_email = new helper.Email('test@example.com');
var subject = 'Hello World from the SendGrid Node.js Library!';
var content = new helper.Content('text/plain', 'Hello, Email!');
var mail = new helper.Mail(from_email, subject, to_email, content);

var sg = require('sendgrid')(process.env.SENDGRID_API_KEY);
var request = sg.emptyRequest({
  method: 'POST',
  path: '/v3/mail/send',
  body: mail.toJSON(),
});

sg.API(request, function(error, response) {
  console.log(response.statusCode);
  console.log(response.body);
  console.log(response.headers);
});
```

One way to improve it could be:

```javascript
var mail = require('sendgrid').mail;
mail.AddFromEmail('test@example.com');
mail.AddToEmail('test@example.com');
mail.AddSubject('Hello World from the SendGrid Node.js Library!');
mail.AddContent('text/plain', 'Hello Email!');

var sendgrid = require('sendgrid')(process.env.SENDGRID_API_KEY);
sendgrid.send(mail, function(error, response) {
[...]

PR #327, opened 2016-10-22
Send single mail

Puigcerber:
Hi @thinkingserious! I've found some time to start working on #312 but before keep going I want your opinion because I wiped out the mail helper to start fresh with a class that simplifies the API.

There was only two tests for the mail helper and with these changes the first one with the minimum to send a single mail pass. So if you consider this is the way to go I can work on the Mail class to pass the second test.

### DIFF
--- a/lib/helpers/mail/mail.js
+++ b/lib/helpers/mail/mail.js
@@ -1,929 +1,283 @@
 'use strict';
-// This helper builds the request body for the v3 mail/send endpoint
 // Please see examples/helpers/mail/example.js for an usage example
 
-function ClickTracking(enable, enable_text) {
-  this.enable = enable;
-  this.enable_text = enable_text;
-
-  this.setEnable = function(enable) {
-    this.enable = enable;
-  };
-
-  this.getEnable = function() {
-    return this.enable;
-  };
-
-  this.setEnableText = function(enable_text) {
-    this.enable_text = enable_text;
-  };
-
-  this.getEnableText = function() {
-    return this.enable_text;
-  };
-
-  this.toJSON = function() {
-    var json = {
-      enable: this.getEnable(),
-      enable_text: this.getEnableText(),
-    };
-    return json;
-  };
-
-  return this;
+function Attachment(content, filename) {
+  this.content = content;
+  this.filename = filename;
 }
 
-function OpenTracking(enable, substitution_tag) {
-  this.enable = enable;
-  this.substitution_tag = substitution_tag;
+Attachment.prototype.setType = function(type) {
+  return this.type = type;
+};
+
+Attachment.prototype.setDisposition = function(disposition) {
+  return this.disposition = disposition;
+};
+
+Attachment.prototype.setContentId = function(content_id) {
+  return this.content_id = content_id;
+};
+
+function MailSettings() { }
+
+MailSettings.prototype.setBcc = function(enable, email) {
+  return this.bcc = {
+    enable: enable,
+    email: email
+  };
+};
+
+MailSettings.prototype.setFooter = function(enable, text, html) {
+  return this.footer = {
+    enable: enable,
+    text: text,
+    html: html
+  };
+};
+
+MailSettings.prototype.setSandBoxMode = function(enable) {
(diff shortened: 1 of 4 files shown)


thinkingserious:
@Puigcerber,

It looks like you are heading in the right direction :)

- Could you please describe how you envision the final call signature will look like? It would be great to get some feedback from the community on that.

- Could you please [sign our CLA](https://github.com/sendgrid/sendgrid-nodejs/blob/master/CONTRIBUTING.md#cla), so we will be able to merge your changes?

Thanks!

lib/helpers/mail/mail.js
```diff
@@ -2,928 +2,48 @@
...
+  this.addFromEmail(from_email);
+  this.addSubject(subject);
+  this.addToEmail(to_email);
+  this.addContent(content);
+}
 
-  this.setReplyTo = function(reply_to) {
-    this.reply_to = reply_to;
-  };
+Mail.prototype.addFromEmail = function (email) {
```
heitortsergent: Just a suggestion, I think it would be good to stick with the previous convention of `setX` method signature for properties that are not arrays, and `addX` for arrays.

heitortsergent: And also, `Email` in the method name might be redundant.

heitortsergent: @thinkingserious I just realized that my suggestions are contrary to what you suggested in #312. 😛 

Is that method signature of `addFromEmail` in line with what the other libraries are using?

Puigcerber: Hi @heitortsergent! I thought about it but followed the suggestions of @thinkingserious. Although I agree with you that we should use `setX` for the from email, subject, etc. This was just a proof of concept. :)

thinkingserious: +1 on the `setX` and `addX`

+1 on `addFrom` instead of `addFromEmail`

Thanks!

Puigcerber:
I've signed and sent the CLA @thinkingserious and hope to find some time this weekend to keep working on the refactor.

Puigcerber:
Hi @thinkingserious @heitortsergent! I've finalized the refactor to pass the second test. Let me know what do you guys think.

heitortsergent:
@Puigcerber nice work! 💥 

I need a little bit more time to go through all the changes, but it looks good to me. One quick question, did you have any specific reason for moving the functions to use `prototype`?

heitortsergent:
@thinkingserious thinking about email best practices and the example you posted on #312, should we add an abstraction to `addTo` and the other methods that take emails so users don't have to pass in a `name`? For example:

`addTo("example@email.com")`
`addTo("example@email.com", "Example Name")`

Are both valid options?

Puigcerber:
Hi @heitortsergent! As I wiped out the previous code and started fresh, I tend to favor prototype methods over constructor methods. If we were to send multiple emails creating instances of the Mail object, memory consumption should be lower with the prototype shared among all the instances.

For the other question I tried to be as flexible as possible, so the following will work:

```js
mail.addTo('example@email.com');
mail.addTo('example@email.com', 'Example Name');
mail.addTo({ email: 'example@email.com', name: 'Example Name'});
```

Puigcerber:
Hey! What's the status on this @thinkingserious?

thinkingserious:
Hi @Puigcerber,

Thanks for checking in! This PR is currently #21 in our backlog. 

It can rise faster in our queue if others will +1 this PR.

SendGridDX:
Hello @Puigcerber,

This issue has finally made it to the top of my queue.

Before I proceed, I'd like to get feedback from @adamreisnz, as he has offered to help us refactor our mail helper based on his repo: https://github.com/adamreisnz/sendgrid-mailer

With Best Regards,

Elmer

adamreisnz:
@SendGridDX @thinkingserious 

And working on the refactoring has nearly reached the top of my queue as well! 
It's time, I will get on it in the next few days! :D

thinkingserious:
This PR is moved to here: https://github.com/sendgrid/sendgrid-nodejs/pull/378

PR #325, opened 2016-10-17
[WIP] Run prism for tests

MikeRalphson:
Refs #316 

I wanted to implement this as I thought I should have working tests before looking at any other issues.

Maybe @tsriram and I could collaborate on this?

Current limitations; does not work on Node 0.10 or 0.12, which are both at or near end-of-life... (as a bonus, it does pass on Node 6, as well as 4 and iojs). 

Please let me know if I'm working along the right lines.

PR #328, opened 2016-10-24
Drop 0.10, 0.12 from supported Node.js versions, add 6

MikeRalphson:
As discussed in #325 with @thinkingserious - PR to drop support for end-of-life Node.js v0.10 as of 01-Nov-2016. Also include Node.js LTS v6 in the list of supported versions, as all tests pass (using Prism).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: node_js
 node_js:
-- "0.10"
-- "0.12"
 - "iojs"
 - "4"
+- "6"
 before_install:
 - npm install -g npm@2
 script:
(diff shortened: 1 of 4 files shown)


heitortsergent:
@thinkingserious looks like this is ready to be merged 👍

thinkingserious:
This one is fairly high on our backlog, but it can climb higher if we can get some more votes on it.

MikeRalphson:
@thinkingserious happy to update this to remove support for Node 0.12, as EOL today.

thinkingserious:
@MikeRalphson,

Great idea, thanks!

Issue #290, opened 2016-09-12
BREAKING CHANGES


thinkingserious:
As a result of your feedback and as a continuation of the execution of our long term roadmap, there will be breaking changes coming soon. The last time we had a major breaking change, it was for the implementation of v3 Web API support. We announced those changes here on GitHub along with some instructions on how to test and provide feedback. 

The feedback we got was amazing, but we didn't quite get the amount or thoroughness of feedback we were hoping for.

We want to continue improving this iterative process, so we are reaching out to you for feedback in order to determine the optimum way to move forward as this library is designed to be community driven, SendGrid led.

**Please take a moment to share your thoughts with us.** 

Following are some ideas we are considering, we will likely choose one from Execution and one from Communication, but not all of the items below.

Execution (for large changes):
1. Utilize the [pre-release functionality](https://help.github.com/articles/creating-releases/) in GitHub to gather feedback and contributions to a specific branch
2. Communicate through a dedicated GitHub issue that references that pre-release branch (e.g [C# Dynamic Update](https://github.com/sendgrid/sendgrid-csharp/issues/303))

Communication:
1. Continue using GitHub issues
2. Opt-in email newsletter for announcements
3. Opt-in email mailing list
4. Dedicated social media account (e.g. @sendgrid_libraries)

How do you prefer to get these announcements? Is this too much? Too little? Please let us know!! 

As always, we are listening carefully and are looking forward to working with you.

**Note: We will always follow the semver.org convention of using major point releases to signify breaking changes. Please DO NOT auto-update your dependencies. It is important to take a look at the CHANGELOG or releases to find out how the breaking changes will impact your code.**

IonicaBizau:
> Utilize the pre-release functionality in GitHub to gather feedback and contributions to a specific branch

That's cool, but in addition to it, I'd suggest `npm publish`ing versions like `3.0.0-alpha1`, `3.0.0-alpha2`, `3.0.0-beta` etc—so we can easily `npm install` them. 🚀

adambuczynski:
Hi @thinkingserious , I would vote for using the pre-release functionality, and where it makes sense, simply keep communication about a specific item in either the issue or the related PR.

[...]

PR #378, opened 2017-04-04
API Simplification

adamreisnz:
# Mail service for the Sendgrid API
This is a dedicated service for interaction with the mail endpoint of the Sendgrid API.

### Initialization with API key
Load the library and set the API key if you haven’t set it before:

```js
//Load library
const sgMail = require('@sendgrid/mail');

//Set API key
sgMail.setApiKey(process.env.SENDGRID_API_KEY);
```

### Single email to one recipient
Load the library, prepare your email data and use the `send` method:

```js
//Load library
const sgMail = require('@sendgrid/mail');

//Create email data
const data = {
  to: 'recipient@example.org',
  from: 'sender@example.org',
  subject: 'Hello world',
  text: 'Hello plain world!',
  html: '<p>Hello HTML world!</p>',
};

//Send email
sgMail.send(data);
```

### Single email to multiple recipients
The `to` field can contain an array of recipients, which will send a single email with all of the recipients in the `to` field. The recipients will be able to see each other:

```js
//Load library
const sgMail = require('@sendgrid/mail');

//Create email data
const data = {
  to: ['recipient1@example.org', 'recipient2@example.org'],
  from: 'sender@example.org',
  subject: 'Hello world',
  text: 'Hello plain world!',
  html: '<p>Hello HTML world!</p>',
};

//Send email
sgMail.send(data);
```

### Multiple emails to multiple recipients
If you want to send multiple _individual_ emails to multiple recipient where they don't see each others email addresses, use `sendMultiple` instead:

```js
//Load library
const sgMail = require('@sendgrid/mail');

//Create email data
const data = {
  to: ['recipient1@example.org', 'recipient2@example.org'],
  from: 'sender@example.org',
  subject: 'Hello world',
  text: 'Hello plain world!',
  html: '<p>Hello HTML world!</p>',
};

//Send emails
sgMail.sendMultiple(data);
```

Note that `sendMultiple(data)` is a convenience shortcut for `send(data, true)`.

### Multiple single emails
The `send` method also accepts an array of email data if you want to send multiple different single emails with for example different content and sender values. This will send multiple requests (in parallel), so be aware of any API rate restrictions:

```js
//Load library
const sgMail = require('@sendgrid/mail');

//Create emails data
const emails = [
  {
    to: 'recipient1@example.org',
    from: 'sender@example.org',
    subject: 'Hello recipient 1',
    text: 'Hello plain world!',
    html: '<p>Hello HTML world!</p>',
  },
  {
    to: 'recipient2@example.org',
    from: 'other-sender@example.org',
    subject: 'Hello recipient 2',
    text: 'Hello other plain world!',
    html: '<p>Hello other HTML world!</p>',
  },
];

//Send emails
sgMail.send(emails);
```

### Flexible email address fields
The email address fields (`to`, `from`, `cc`, `bcc`, `replyTo`) are flexible and can be any of the following:

```js
const data = {

  //Simple email address string
  to: 'someone@example.org',

  //Email address with name
  to: 'Some One <someone@example.org>',

  //Object with name/email
  to: {
    name: 'Some One',
    email: 'someone@example.org',
  },
};
```

### Handling success/failure
The `send` and `sendMultiple` methods return a `Promise`, so you can handle success and capture errors:

```js
sgMail
  .send(data)
  .then(() => {
    //Celebrate
  })
  .catch(error => {
    //Do something with the error
  });
```

Alternatively, pass a callback function as the last parameter:

```js
sgMail
  .send(data, (error, result) => {
    if (error) {
      //Do something with the error
    }
    else {
      //Celebrate
    }
  });
```

## Advanced usage
All other advanced settings are supported and can be passed in through the data object according to the expected format as per the [API v3 documentation](https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/index.html). Note that you can use either `camelCase` or `snake_case` for property names.

### Using transactional templates
Configure the substitution tag wrappers:

```js
sgMail.setSubstitutionWrappers('{{', '}}');
```

Then provide a template ID and substitutions:

```js
const data = {
  templateId: 'sendgrid-template-id',
  substitutions: {
    name: 'Some One',
    id: '123',
  },
};
```

### Customization per recipient
To send multiple individual emails to multiple recipients with a different subject and/or substitutions, expand the `to` array as follows:

```js
const data = {
  to: [
    {
      email: 'recipient1@example.org',
      subject: 'Hello recipient 1',
      substitutions: {
        name: 'Recipient 1',
        id: '123',
      },
    },
    {
      email: 'recipient2@example.org',
      subject: 'Hello recipient 2',
      substitutions: {
        name: 'Recipient 2',
        id: '456',
      },
    }
  ],
  from: 'sender@example.org',
  text: 'Hello plain world!',
  html: '<p>Hello HTML world!</p>',
};
```

### Sending attachments
Attachments can be sent by providing an array of `attachments` as per the API specifications:

```js
const data = {
  to: 'recipient@example.org',
  from: 'sender@example.org',
  subject: 'Hello attachment',
  html: '<p>Here’s an attachment for you!</p>',
  attachments: [
    {
      content: 'Some attachment content',
      filename: 'some-attachment.txt',
    },
  ],
};
```

### Manually providing personalizations
Instead of using the `to` shorthand proper, you can still manually provide `personalizations` as per the API definition:

```js
const data = {
  from: 'sender@example.org',
  text: 'Hello plain world!',
  html: '<p>Hello HTML world!</p>',
  personalizations: [
    {
      to: [
        {
          name: 'Someone',
          email: 'someone@example.org',
        },
      ],
      cc: [
        {
          name: 'Someone Else',
          email: 'someone.else@example.org',
        },
      ],
      subject: 'Some subject',
    }
  ],
};
```

### Manually providing content
Instead of using the `text` and `html` shorthand properties, you can manually use the `content` property:

```js
const data = {
  to: 'recipient@example.org',
  from: 'sender@example.org',
  subject: 'Hello manual content',
  content: [
    {
      type: 'text/html',
      value: '<p>Hello HTML world!</p>',
    },
    {
      type: 'text/plain',
      value: 'Hello plain world!',
    },
  ],
};
```

### Specifying time to send at
Use the `sendAt` property to specify when to send the emails (in seconds, not milliseconds):

```js
const data = {
  to: 'recipient@example.org',
  from: 'sender@example.org',
  subject: 'Hello delayed email',
  html: '<p>Some email content</p>',
  sendAt: 1500077141,
};
```

### Specifying custom headers
Use the `headers` property to specify any custom headers:

```js
const data = {
  to: 'recipient@example.org',
  from: 'sender@example.org',
  subject: 'Hello custom header',
  html: '<p>Some email content</p>',
  headers: {
    'X-CustomHeader': 'Custom header value',
  },
};
```

### Specifying categories
Use the `categories` property to provide an array of categories for your email:

```js
const data = {
  to: 'recipient@example.org',
  from: 'sender@example.org',
  subject: 'Hello email with categories',
  html: '<p>Some email content</p>',
  categories: [
    'transactional', 'customer', 'weekly',
  ],
};
```

### Other options
All other options from the [API definition](https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/index.html) are supported:

```js
const data = {

  //Sections
  sections: {},

  //Custom arguments
  customArgs: {},

  //Batch ID
  batchId: String,

  //ASM
  asm: {},

  //IP pool name
  ipPoolName: String,

  //Mail settings
  mailSettings: {},

  //Tracking settings
  trackingSettings: {},
}
```

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+node_modules
+examples
+test
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,0 +1,136 @@
+root: true
+extends: eslint:recommended
+parserOptions:
+  ecmaVersion: 6
+  ecmaFeatures:
+    impliedStrict: true
+  sourceType: module
+env:
+  es6: true
+  node: true
+  mocha: true
+  jasmine: true
+rules:
+  linebreak-style:
+    - error
+    - unix
+  max-len:
+    - warn
+    - code: 80
+      tabWidth: 2
+      ignoreComments: true
+      ignoreUrls: true
+  indent:
+    - error
+    - 2
+    - SwitchCase: 1
+      VariableDeclarator:
+        let: 2
+        const: 3
+  semi:
+    - error
+    - always
+  consistent-this:
+    - error
+    - self
+    - $ctrl
+  quotes:
+    - error
+    - single
+    - allowTemplateLiterals: true
+  curly:
+    - error
+    - all
+  comma-dangle:
+    - error
+    - always-multiline
+  new-cap:
+    - error
+    - newIsCap: true
+      capIsNew: true
+      properties: false
+  array-bracket-spacing:
+    - error
+    - never
+  arrow-spacing:
+    - error
+    - before: true
+      after: true
+  block-spacing:
+    - error
+    - always
+  comma-spacing:
+    - error
+    - before: false
+      after: true
+  computed-property-spacing:
+    - error
+    - never
+  generator-star-spacing:
+    - error
+    - before: true
+      after: false
+  key-spacing:
+    - error
+    - beforeColon: false
+      afterColon: true
+      mode: minimum
+  keyword-spacing:
+    - error
+    - before: true
+  semi-spacing:
+    - error
+    - before: false
+      after: true
+  space-in-parens:
+    - error
+    - never
+  space-unary-ops:
+    - error
+    - words: true
+      nonwords: false
+  space-before-function-paren:
+    - error
+    - never
+  space-before-blocks:
+    - error
+    - always
+  yoda:
+    - error
+    - never
+  wrap-iife:
+    - error
+    - outside
+  eqeqeq:
+    - error
+    - always
+  newline-per-chained-call:
+    - error
+    - ignoreChainWithDepth: 3
+  one-var-declaration-per-line:
+    - error
+    - initializations
+  brace-style:
+    - error
+    - stroustrup
+  no-implicit-coercion:
+    - error
+    - boolean: false
+  no-multiple-empty-lines:
+    - error
+    - max: 1
+  eol-last: error
+  dot-notation: error
+  space-infix-ops: error
+  no-with: error
+  no-unreachable: error
+  no-redeclare: error
+  no-unexpected-multiline: error
+  no-multi-spaces: error
+  no-multi-str: error
+  no-trailing-spaces: error
+  no-mixed-spaces-and-tabs: error
+  no-spaced-func: error
+  no-whitespace-before-property: error
+  no-lonely-if: error
+  no-console: off
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ test/config.js
 .env.staging
 .env
 temp.js
+.vscode/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,15 @@
 language: node_js
 node_js:
-- "iojs"
 - "4"
 - "6"
+- "7"
 before_install:
 - npm install -g npm@2
+before_script:
+- mkdir prism
+- mkdir prism/bin
+- export PATH=$PATH:$PWD/prism/bin/
+- ./test/prism.sh
 script:
 - npm test
 - npm run test:typescript
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,13 @@
 language: node_js
 node_js:
-- "0.10"
-- "0.12"
 - "iojs"
 - "4"
+- "6"
 before_install:
 - npm install -g npm@2
+script:
+- npm test
+- npm run test:typescript
 notifications:
   hipchat:
     rooms:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,144 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+## [5.0] - 2017-03-31 ##
+### BREAKING CHANGE
+- Pull #328
+- Drop 0.10, 0.12 from supported Node.js versions, add 6
+- Thanks to [Mike Ralphson](https://github.com/MikeRalphson) for the PR!
+
+## [4.10] - 2017-03-30 ##
+### Added
+- Pull #303
+- Add Inbound Parse data parser
+- Thanks to [Jamie](https://github.com/jamsinclair) for the PR!
+
+## [4.9] - 2017-03-28 ##
+### Added
+- Pull #371
+- Catch error in sendgrid, partially solves #370
+- Thanks to [gf](https://github.com/furstenheim) for the PR!
+
+## [4.8.4] - 2017-03-14 ##
+### Fixed
+- Pull #348
+- Fix typescript definition for the ClickTracking
+- Thanks to [dhenriques](https://github.com/dhenriques) for the PR!
+
+## [4.8.3] - 2017-03-14 ##
+### Fixed
+- Pull #368, Fixes #367
+- Personalization.getSubstitutions() return type incorrect in index.d.ts TypeScript definition file
+- Thanks to [Brian Love](https://github.com/blove) for the PR!
+
+## [4.8.2] - 2017-03-14 ##
+### Fixed
+- Pull #338
+- Fix function name in mail helper and fix the corresponding test
+- Thanks to [Seoker Wang](https://github.com/seoker) for the PR!
+
+## [4.8.1] - 2017-03-13 ##
+### Fixed
+- Pull #354
+- fix: Make various TypeScript request components optional
+- Thanks to [Benjamin Pannell](https://github.com/SPARTAN563) for the PR!
+
+## [4.8.0] - 2017-03-10 ##
+### Added
+- Pull #333
+- [TypeDefinition] SendGrid.API should return Promise<T> instead of PromiseLike<T>
+- Thanks to [Ayman Nedjmeddine](https://github.com/IOAyman) for the PR!
+
+## [4.7.1] - 2016-10-25 ##
+### Added
+- Pull #329
+- Simplified installation instructions
+- Thanks to [Heitor Tashiro Sergent](https://github.com/heitortsergent) for the PR!
+
+## [4.7.0] - 2016-10-14 ##
+### Added
+- Pull #323, Fixed issue #317
+- Updates nodejs-http-client dependency to [v2.3.0](https://github.com/sendgrid/nodejs-http-client/releases/tag/v2.3.0)
+- Invoke the API callback with a mocked response upon Error
+- Thanks to [Huli](https://github.com/aszx87410) for the PR!
+
+## [4.6.0] - 2016-10-13 ##
+### Added
+- Pull #319, Fixed issue #266
+- Converts `response.body` to have valid json objects
+- Thanks to [Hugo Durães](https://github.com/hugoduraes) for the pull request!
+
+## [4.5.0] - 2016-10-05 ##
+## Added
+- Pull #308 [feat: Add support for promises to the TypeScript definitions file](https://github.com/sendgrid/sendgrid-nodejs/pull/308)
+- Thanks to [Benjamin Pannell](https://github.com/SPARTAN563) for the pull request!
+
+## [4.4.1] - 2016-09-27 ##
+## Fixed
+- Pull #309 [Allow emptyRequest() to be called with a request object](https://github.com/sendgrid/sendgrid-nodejs/pull/309)
+- Thanks to [Benjamin Pannell](https://github.com/SPARTAN563) for the fix!
+
+## [4.4.0] - 2016-09-27 ##
+## Added
+- Pull #299: [Fill 'response' property in error instance](https://github.com/sendgrid/sendgrid-nodejs/pull/299)
+- This makes the behavior when using callbacks consistent with the promise counterpart.
+That is, in both cases, error.response is present
+- Thanks to [Guilherme Souza](https://github.com/sitegui) for the pull request!
+
+## [4.3.1] - 2016-09-27 ##
+## Fixed
+- Pull #300: [Correct TypeScript definitions](https://github.com/sendgrid/sendgrid-nodejs/pull/300)
+- Thanks to [Benjamin Pannell](https://github.com/SPARTAN563) for the fix!
+
+## [4.3.0] - 2016-09-15 ##
+## Added
+- Pull #251: [Add TypeScript typings definition file for v3 API](https://github.com/sendgrid/sendgrid-nodejs/pull/251)
+- Thanks to [Benjamin Pannell](https://github.com/SPARTAN563) for the pull request!
+
+## [4.2.1] - 2016-09-15 ##
+### Fixed
+- Issue #292: [4.2.0 Error: Cannot find module 'debug'](https://github.com/sendgrid/sendgrid-nodejs/issues/292)
+
+## [4.2.0] - 2016-09-14 ##
+### Added
+- Pull Request #278: [Tkp/contact importer](https://github.com/sendgrid/sendgrid-nodejs/pull/278)
+- Thanks to [Tom Kirkpatrick](https://github.com/mrfelton) for the pull request!
+
+## [4.1.0] - 2016-09-09 ##
+### Added
+- Pull Request #277: [Make SendGrid a factory that enables multiple prototype instances](https://github.com/sendgrid/sendgrid-nodejs/pull/277)
+- Thanks to [Tom Kirkpatrick](https://github.com/mrfelton) for the pull request!
+
+## [4.0.2] - 2016-08-24 ##
+### Added
+- Table of Contents in the README
+- Added a [USE_CASES.md](https://github.com/sendgrid/sendgrid-nodejs/blob/master/USE_CASES.md) section, with the first use case example for transactional templates
+
+## [4.0.1] - 2016-08-02 ##
+### Fixed
+- Pull request [#264](https://github.com/sendgrid/sendgrid-nodejs/pull/264): Fixed backwards compatability with Node.js versions 0.10 and 0.12
+- Use var instead of let
+- Check if Promise is defined
+
+## [4.0.0] - 2016-08-02 ##
+### Breaking Change
+- Pull request [#261](https://github.com/sendgrid/sendgrid-nodejs/pull/261)
+- BIG THANKS to [Adam Buczynski](https://github.com/adambuczynski)!!!
+
+### Fixed
+- Issue [#240](https://github.com/sendgrid/sendgrid-nodejs/issues/240)
+- Issue [#246](https://github.com/sendgrid/sendgrid-nodejs/issues/246)
+- Issue [#206](https://github.com/sendgrid/sendgrid-nodejs/issues/206)
+
+### Added
+- Extracted some logic into helpers
+- Using a getEmptyRequest helper to avoid code duplication
+- emtpyRequest now accepts an object with data to extend the empty request with, this will allow simpler syntax for initializing requests.
+- Callback function now receives two parameters as per Node conventions (error, response)
+- If no callback provided, the method will return a promise instead.
+- Implemented promise API when not passing a callback function
+- Using native Promise by default if present, but allow users to override this with any other implementation by setting Sendgrid.Promise to any value, e.g. Sendgrid.Promise = require('bluebird')
+
 ## [3.0.11] - 2016-07-26 ##
 ### Added
 - [Troubleshooting](https://github.com/sendgrid/sendgrid-python/blob/master/TROUBLESHOOTING.md) section
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,19 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+## [5.2.0] - 2017-05-3 ##
+### Added
+- Pull #325
+- Run prism for tests
+- Now the mock SendGrid server, powered by [Prism](https://stoplight.io/platform/prism/) is automated locally and on Travis
+- Thanks to [Mike Ralphson](https://github.com/MikeRalphson) for the PR!
+
+## [5.0.1] - 2017-04-24 ##
+### Fixed
+- Pull #385
+- Remove JSON.parse() around response bodies in contact-importer
+- Thanks to [Eemeli Aro](https://github.com/eemeli) for the PR!
+
 ## [5.0] - 2017-03-31 ##
 ### BREAKING CHANGE
 - Pull #328
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,13 +16,13 @@ We use [Milestones](https://github.com/sendgrid/sendgrid-nodejs/milestones) to h
 <a name="cla"></a>
 ## CLAs and CCLAs
 
-Before you get started, SendGrid requires that a SendGrid Contributor License Agreement (CLA) or a SendGrid Company Contributor Licensing Agreement (CCLA) be filled out by every contributor to a SendGrid open source project.
+Before you get started, SendGrid requires that a SendGrid Contributor License Agreement (CLA) be filled out by every contributor to a SendGrid open source project.
 
-Our goal with the CLA and CCLA is to clarify the rights of our contributors and reduce other risks arising from inappropriate contributions.  The CLA also clarifies the rights SendGrid holds in each contribution and helps to avoid misunderstandings over what rights each contributor is required to grant to SendGrid when making a contribution.  In this way the CLA and CCLA encourage broad participation by our open source community and help us build strong open source projects, free from any individual contributor withholding or revoking rights to any contribution.
+Our goal with the CLA is to clarify the rights of our contributors and reduce other risks arising from inappropriate contributions.  The CLA also clarifies the rights SendGrid holds in each contribution and helps to avoid misunderstandings over what rights each contributor is required to grant to SendGrid when making a contribution.  In this way the CLA encourages broad participation by our open source community and helps us build strong open source projects, free from any individual contributor withholding or revoking rights to any contribution.
 
-SendGrid does not merge a pull request made against a SendGrid open source project until that pull request is associated with a signed CLA (or CCLA). Copies of the CLA and CCLA are available [here](https://drive.google.com/a/sendgrid.com/file/d/0B0PlcM9qA91LN2VEUTJWU2RIVXc/view).
+SendGrid does not merge a pull request made against a SendGrid open source project until that pull request is associated with a signed CLA. Copies of the CLA are available [here](https://gist.github.com/SendGridDX/98b42c0a5d500058357b80278fde3be8#file-sendgrid_cla).
 
-You may submit your completed [CLA or CCLA](https://drive.google.com/a/sendgrid.com/file/d/0B0PlcM9qA91LN2VEUTJWU2RIVXc/view) to SendGrid at [dx@sendgrid.com](mailto:dx@sendgrid.com).  SendGrid will then confirm you are ready to begin making contributions.
+When you create a Pull Request, after a few seconds, a comment will appear with a link to the CLA. Click the link and fill out the brief form and then click the "I agree" button and you are all set. You will not be asked to re-sign the CLA unless we make a change.
 
 There are a few ways to contribute, which we'll enumerate below:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ Before you decide to create a new issue, please try the following:
 
 ### Please use our Bug Report Template
 
-In order to make the process easier, we've included a [sample bug report template](https://github.com/sendgrid/sendgrid-nodejs/.github/ISSUE_TEMPLATE) (borrowed from [Ghost](https://github.com/TryGhost/Ghost/)). The template uses [GitHub flavored markdown](https://help.github.com/articles/github-flavored-markdown/) for formatting.
+In order to make the process easier, we've included a [sample bug report template](https://github.com/sendgrid/sendgrid-nodejs/blob/master/.github/ISSUE_TEMPLATE) (borrowed from [Ghost](https://github.com/TryGhost/Ghost/)). The template uses [GitHub flavored markdown](https://help.github.com/articles/github-flavored-markdown/) for formatting.
 
 <a name="improvements_to_the_codebase"></a>
 ## Improvements to the Codebase
@@ -64,7 +64,7 @@ We welcome direct contributions to the sendgrid-nodejs code base. Thank you!
 
 ##### Prerequisites #####
 
-- Node.js version 0.10, 0.12 or 4
+- Node.js version 4 or 6
 - Please see [package.json](https://github.com/sendgrid/sendgrid-nodejs/tree/master/package.json)
 
 ##### Initial setup: #####
@@ -72,6 +72,7 @@ We welcome direct contributions to the sendgrid-nodejs code base. Thank you!
 ```bash
 git clone https://github.com/sendgrid/sendgrid-nodejs.git
 cd sendgrid-nodejs
+npm install
 ```
 
 ## Environment Variables
@@ -129,6 +130,16 @@ All test files are in the [`tests`](https://github.com/sendgrid/sendgrid-nodejs/
 For the purposes of contributing to this repo, please update the [`test.js`](https://github.com/sendgrid/sendgrid-nodejs/tree/master/test/test.js) file with unit tests as you modify the code.
 
 To run the tests:
+
+First, we need to setup a local mocked SendGrid API server.
+
+Install StopLight.io's [Prism](https://stoplight.io/platform/prism/).
+
+```bash
+./prism run --mock --list --spec https://raw.githubusercontent.com/sendgrid/sendgrid-oai/master/oai_stoplight.json
+```
+
+Now we can run the tests at the root of the `sendgrid-nodejs` repo:
 
 ```bash
 mocha
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,7 @@ We welcome direct contributions to the sendgrid-nodejs code base. Thank you!
 
 ##### Prerequisites #####
 
-- Node.js version 4 or 6
+- Node.js version 4, 6 or 7
 - Please see [package.json](https://github.com/sendgrid/sendgrid-nodejs/tree/master/package.json)
 
 ##### Initial setup: #####
@@ -129,17 +129,7 @@ All test files are in the [`tests`](https://github.com/sendgrid/sendgrid-nodejs/
 
 For the purposes of contributing to this repo, please update the [`test.js`](https://github.com/sendgrid/sendgrid-nodejs/tree/master/test/test.js) file with unit tests as you modify the code.
 
-To run the tests:
-
-First, we need to setup a local mocked SendGrid API server.
-
-Install StopLight.io's [Prism](https://stoplight.io/platform/prism/).
-
-```bash
-./prism run --mock --list --spec https://raw.githubusercontent.com/sendgrid/sendgrid-oai/master/oai_stoplight.json
-```
-
-Now we can run the tests at the root of the `sendgrid-nodejs` repo:
+Run the tests at the root of the `sendgrid-nodejs` repo:
 
 ```bash
 mocha
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![BuildStatus](https://travis-ci.org/sendgrid/sendgrid-nodejs.svg?branch=master)](https://travis-ci.org/sendgrid/sendgrid-nodejs)
+[![npm version](https://badge.fury.io/js/sendgrid.svg)](https://badge.fury.io/js/sendgrid)
 
 Please see our announcement regarding [breaking changes](https://github.com/sendgrid/sendgrid-nodejs/issues/290). Your support is appreciated!
 
@@ -29,7 +30,7 @@ We appreciate your continued support, thank you!
 
 ## Prerequisites
 
-- Node.js version 4 or 6
+- Node.js version 4, 6 or 7
 - The SendGrid service, starting at the [free level](https://sendgrid.com/free?source=sendgrid-nodejs)
 
 ## Setup Environment Variables
@@ -65,20 +66,23 @@ The following is the minimum needed code to send an email with the [/mail/send H
 
 ```javascript
 var helper = require('sendgrid').mail;
-var from_email = new helper.Email('test@example.com');
-var to_email = new helper.Email('test@example.com');
-var subject = 'Hello World from the SendGrid Node.js Library!';
-var content = new helper.Content('text/plain', 'Hello, Email!');
-var mail = new helper.Mail(from_email, subject, to_email, content);
+var fromEmail = new helper.Email('test@example.com');
+var toEmail = new helper.Email('test@example.com');
+var subject = 'Sending with SendGrid is Fun';
+var content = new helper.Content('text/plain', 'and easy to do anywhere, even with Node.js');
+var mail = new helper.Mail(fromEmail, subject, toEmail, content);
 
 var sg = require('sendgrid')(process.env.SENDGRID_API_KEY);
 var request = sg.emptyRequest({
   method: 'POST',
   path: '/v3/mail/send',
-  body: mail.toJSON(),
+  body: mail.toJSON()
 });
 
-sg.API(request, function(error, response) {
+sg.API(request, function (error, response) {
+  if (error) {
+    console.log('Error response received');
+  }
   console.log(response.statusCode);
   console.log(response.body);
   console.log(response.headers);
@@ -101,39 +105,39 @@ var request = sg.emptyRequest({
       {
         to: [
           {
-            email: 'test@example.com',
-          },
+            email: 'test@example.com'
+          }
         ],
-        subject: 'Hello World from the SendGrid Node.js Library!',
-      },
+        subject: 'Sending with SendGrid is Fun'
+      }
     ],
     from: {
-      email: 'test@example.com',
+      email: 'test@example.com'
     },
     content: [
       {
         type: 'text/plain',
-        value: 'Hello, Email!',
-      },
-    ],
-  },
+        value: 'and easy to do anywhere, even with Node.js'
+      }
+    ]
+  }
 });
 
-//With promise
+// With promise
 sg.API(request)
-  .then(response => {
+  .then(function (response) {
     console.log(response.statusCode);
     console.log(response.body);
     console.log(response.headers);
   })
-  .catch(error => {
-    //error is an instance of SendGridError
-    //The full response is attached to error.response
+  .catch(function (error) {
+    // error is an instance of SendGridError
+    // The full response is attached to error.response
     console.log(error.response.statusCode);
   });
 
-//With callback
-sg.API(request, function(error, response) {
+// With callback
+sg.API(request, function (error, response) {
   if (error) {
     console.log('Error response received');
   }
@@ -146,7 +150,7 @@ sg.API(request, function(error, response) {
 ## General v3 Web API Usage
 
 ```javascript
-var sg = require('sendgrid')(process.env.SENDGRID_API_KEY)
+var sg = require('sendgrid')(process.env.SENDGRID_API_KEY);
 
 // GET Collection
 var request = sg.emptyRequest({
@@ -154,28 +158,28 @@ var request = sg.emptyRequest({
   path: '/v3/api_keys'
 });
 
-//With promise
+// With promise
 sg.API(request)
-  .then(response => {
-    console.log(response.statusCode)
-    console.log(response.body)
-    console.log(response.headers)
+  .then(function (response) {
+    console.log(response.statusCode);
+    console.log(response.body);
+    console.log(response.headers);
   })
-  .catch(error => {
-    //error is an instance of SendGridError
-    //The full response is attached to error.response
+  .catch(function (error) {
+    // error is an instance of SendGridError
+    // The full response is attached to error.response
     console.log(error.response.statusCode);
   });
 
-//With callback
-sg.API(request, function(error, response) {
+// With callback
+sg.API(request, function (error, response) {
   if (error) {
     console.log('Error response received');
   }
-  console.log(response.statusCode)
-  console.log(response.body)
-  console.log(response.headers)
-})
+  console.log(response.statusCode);
+  console.log(response.body);
+  console.log(response.headers);
+});
 ```
 
 <a name="usage"></a>
@@ -187,7 +191,7 @@ sg.API(request, function(error, response) {
 - [How-to: Migration from v2 to v3](https://sendgrid.com/docs/Classroom/Send/v3_Mail_Send/how_to_migrate_from_v2_to_v3_mail_send.html)
 - [v3 Web API Mail Send Helper](https://github.com/sendgrid/sendgrid-nodejs/tree/master/lib/helpers/mail/README.md)
 
-<a name="use_cases">
+<a name="use_cases"></a>
 # Use Cases
 
 [Examples of common API use cases](https://github.com/sendgrid/sendgrid-nodejs/blob/master/USE_CASES.md), such as how to send an email with a transactional template.
@@ -225,5 +229,4 @@ sendgrid-nodejs is guided and supported by the SendGrid [Developer Experience Te
 
 sendgrid-nodejs is maintained and funded by SendGrid, Inc. The names and logos for sendgrid-nodejs are trademarks of SendGrid, Inc.
 
-![SendGrid Logo]
-(https://uiux.s3.amazonaws.com/2016-logos/email-logo%402x.png)
+![SendGrid Logo](https://uiux.s3.amazonaws.com/2016-logos/email-logo%402x.png)
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
 [![BuildStatus](https://travis-ci.org/sendgrid/sendgrid-nodejs.svg?branch=master)](https://travis-ci.org/sendgrid/sendgrid-nodejs)
 
+Please see our announcement regarding [breaking changes](https://github.com/sendgrid/sendgrid-nodejs/issues/290). Your support is appreciated!
+
 **This library allows you to quickly and easily use the SendGrid Web API v3 via Node.js.**
 
-Version 3.X.X of this library provides full support for all SendGrid [Web API v3](https://sendgrid.com/docs/API_Reference/Web_API_v3/index.html) endpoints, including the new [v3 /mail/send](https://sendgrid.com/blog/introducing-v3mailsend-sendgrids-new-mail-endpoint).
+Version 3.X.X+ of this library provides full support for all SendGrid [Web API v3](https://sendgrid.com/docs/API_Reference/Web_API_v3/index.html) endpoints, including the new [v3 /mail/send](https://sendgrid.com/blog/introducing-v3mailsend-sendgrids-new-mail-endpoint).
 
 This library represents the beginning of a new path for SendGrid. We want this library to be community driven and SendGrid led. We need your help to realize this goal. To help make sure we are building the right things in the right order, we ask that you create [issues](https://github.com/sendgrid/sendgrid-nodejs/issues) and [pull requests](https://github.com/sendgrid/sendgrid-nodejs/blob/master/CONTRIBUTING.md) or simply upvote or comment on existing issues or pull requests.
 
@@ -10,11 +12,24 @@ Please browse the rest of this README for further detail.
 
 We appreciate your continued support, thank you!
 
+# Table of Contents
+
+* [Installation](#installation)
+* [Quick Start](#quick_start)
+* [Usage](#usage)
+* [Use Cases](#use_cases)
+* [Announcements](#announcements)
+* [Roadmap](#roadmap)
+* [How to Contribute](#contribute)
+* [Troubleshooting](#troubleshooting)
+* [About](#about)
+
+<a name="installation"></a>
 # Installation
 
 ## Prerequisites
 
-- Node.js version 0.10, 0.12 or 4
+- Node.js version 4 or 6
 - The SendGrid service, starting at the [free level](https://sendgrid.com/free?source=sendgrid-nodejs)
 
 ## Setup Environment Variables
@@ -31,36 +46,15 @@ source ./sendgrid.env
 
 The following recommended installation requires [npm](https://npmjs.org/). If you are unfamiliar with npm, see the [npm docs](https://npmjs.org/doc/). Npm comes installed with Node.js since node version 0.8.x therefore you likely already have it.
 
-Add the following to your `package.json` file:
-
-```json
-{
-  ...
-  "dependencies": {
-    ...
-    "sendgrid": "^3.0.11"
-  }
-}
-```
-
-Install sendgrid-nodejs and its dependencies:
-
 ```bash
-npm install
-```
-
-### Alternative Installation
-
-You can also install sendgrid locally with the following command:
-
-```bash
-npm install sendgrid
+npm install --save sendgrid
 ```
 
 ## Dependencies
 
 - [Nodejs-HTTP-Client](https://github.com/sendgrid/nodejs-http-client)
 
+<a name="quick_start"></a>
 # Quick Start
 
 ## Hello Email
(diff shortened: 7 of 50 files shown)


SendGridDX:
[CLA assistant check](https://cla.sendgrid.com/sendgrid/sendgrid-nodejs?pullRequest=378) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.sendgrid.com/sendgrid/sendgrid-nodejs?pullRequest=378) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.sendgrid.com/check/sendgrid/sendgrid-nodejs?pullRequest=378) it.</sub>

adamreisnz:
@thinkingserious I've created the above proposal for the 4 use cases required. Mostly this is adapted from my `sendgrid-mailer` library, as it seemed to be convenient and sensible and the community had favourable thoughts about it.

Additional observations:

- The `sendgrid.mail` namespace represents that you are accessing methods in the `mail` endpoint of the API. This will allow us to develop additional namespaces for the library to cover other endpoints as well, and to write convenience helper methods for those in due time.
- I want to keep the helper classes, but refactor them slightly and make them easier to access and use. The above helper functions will use these helper classes internally to properly construct the underlying JSON payload and exposing them to the users will allow for more flexibility if needed.
- All underlying "raw" methods will be made available as well, so you could still send completely custom JSON requests to the API endpoint.
- As there will be some breaking changes this will likely be a new major release with significant refactoring.

I've already signed the CLA, but my Github username changed a while ago, so that might need to be updated in your system (used to be `adambuczynski`).

Let me know your thoughts and how you'd like to open this up for review by the community. Keen to start working on it once it's been approved.

SendGridDX:
Hello @adamreisnz!

I am in agreement with your design.

I believe those interested in this redesign should follow along here. If it becomes difficult to communicate, we can explore other options.

I have alerted people that may be interested in this re-design.

SendGridDX:
@adamreisnz,

With regards to the CLA, we rolled out a new, greatly simplified process. All you need to do now if fill out a short form and click a button. Please do that [here](https://cla.sendgrid.com/sendgrid/sendgrid-nodejs?pullRequest=378). Thanks!

thebigredgeek:
Would be great if this was ironed out quickly.  I know several startups here in Silicon Valley who have moved away from sendgrid because of the new API design :(.  We are also thinking about it

SendGridDX:
Thanks for the feedback @thebigredgeek. 

Do you have any thoughts on the new Mail Helper?

thebigredgeek:
The API looks simple to use.  The question, to me, is how long we will need to wait to start using it.  The new V3 "API" has been very painful for us over the past few months, and has cost us countless man hours trying to debug bizarre sendgrid issues that we never had to deal with before.

SendGridDX:
@thebigredgeek,

Thanks for the additional feedback!

Right now there are too many variables for me to give an accurate release date. 

This particular issue, is at a high priority and we will do all we can to support @adamreisnz to get this ready for a stable release.

With regards to your current issues, please feel free to reach out to our team directly (dx@sendgrid.com) or here on GitHub, we are happy to help.

Also, it would be great to know the particular pain points you ran into so that we don't duplicate them with this next iteration.

thebigredgeek:
@SendGridDX Great!  Sounds like you guys will hopefully ship soon.  The issues with the old V3 "API" were caused by the horrendous design... "phase 1" or not... which required the gluing together of tens to hundreds of custom objects in order to send a single email.

SendGridDX:
Thank you @thebigredgeek once again for the valuable feedback. We will continue to make iterative improvements based on feedback such as yours, including [how we handle releases](https://github.com/sendgrid/sendgrid-nodejs/issues/290). So thanks again for taking the time to express your issues.

In the meantime, we are here to help with any pain you are experiencing right now.

adamreisnz:
@thebigredgeek I'm happy to start working on it soon and get this out quickly. In the mean time, to alleviate your trouble, you can look into using https://www.npmjs.com/package/sendgrid-mailer. It's a wrapper I wrote to simplify the sending of emails with Sendgrid. My proposed API up here is pretty much an adaptation of the API in that package, so moving from it to the new version of the official package shouldn't be very time consuming.

Please let me know if you have any additional feedback on the API or additions you'd like to see to make sending emails for your use cases easier.

thebigredgeek:
Thanks @adamreisnz .  I hope you aren't doing this for free haha.

fisch0920:
Looks awesome! `nodemailer` support?

adamreisnz:
@fisch0920 I might look into that after the new API is there, but the priority at this stage is to release a new and easy to use API

adamreisnz:
@SendGridDX do you want to await more feedback or shall we get going with this?

SendGridDX:
@adamreisnz,

I think we should get going.

Based on the feedback here and other feedback on other issues on your `sendgrid-mailer` package, I think you have laid out a solid path forward.

Thanks!

adamreisnz:
Ok I'll get started. One thing though, I noticed you also added TS support, but I don't want to make my hands dirty with that, as I find that pleasing the TS community is often hard, usually filled with issues, and never very rewarding ;) Can I leave that for you or someone else on the team to implement once done? I say we focus on the pure JS implementation first, as the majority of Node users will still be using and favour ES over TS.

SendGridDX:
@SPARTAN563,

Thoughts on helping us support TS as we update the Mail Helper?

@adamreisnz,

We need to support non End-of-Life versions: https://github.com/nodejs/LTS

adamreisnz:
@SendGridDX Hmm, Node 4 LTS has just entered maintenance phase, so I reckon we can focus on Node 6+ for this new development. Anyone on Node 4 can keep using the existing library and they are probably unlikely to switch to a new major version with a new API anyway. Thoughts?

SPARTAN563:
Hey guys, more than happy to help out - just ping me when you're nearing completion and I'll slot it into my schedule somewhere. If you have a time-box, let me know and I'll try to prioritise to meet that.

SendGridDX:
Hi @adamreisnz,

I not too familiar with what it means to enter maintenance phase. Is there a dramatic drop off in usage when this happens?

And yes, that's true, we can point people to this version of the SDK if they are stuck on Node 4.

adamreisnz:
Well it means there's still one year of maintenance support of that Node version, but this mostly pertains critical security fixes.

> Once a release moves into Maintenance mode, only **critical** bugs, **critical** security fixes, and documentation updates will be permitted.

So personally I don't think at this stage it's necessary to keep offering support for Node 4 with the new API.

SendGridDX:
Does anyone else following this thread have an opinion on supporting (or not) Node.js 4?

thebigredgeek:
I think v4+ is ideal.  Most people don't use old versions of node in production

thinkingserious:
@thebigredgeek,

Thanks for providing quick feedback!

So just supporting v6+ would be too restrictive?

thebigredgeek:
I think v4+ is likely desirable for the foreseeable future.  I think v6+ is likely on the edge.  All this lib is doing is making HTTP requests, so I don't really think we need full ES6 support which is the only real reason I see for cutting v4 compatibility

adamreisnz:
> I think v4+ is likely desirable for the foreseeable future. I think v6+ is likely on the edge. 

I have to disagree here. As you can see in the overview, v4 is only going to be maintained for another year, and v5 is already at the end of its life. v6 is the currently stable and active LTS (been active for over a year), and I am not aware of any projects in my own business environment that's still using v4. Everyone has either migrated to v6 or v7, with v8 around the corner.

So I would say that for the "foreseeable" future we can look past the 1 more year that v4 is going to be supported with critical patches only.

> All this lib is doing is making HTTP requests, so I don't really think we need full ES6 support which is the only real reason I see for cutting v4 compatibility

That's not really true. It's not just about what the library is doing behind the scenes, it's also about ease of development and maintainability. Yes we could go with prototype and write more code, but we can also leverage the ES6 features that make our lives easier like classes, destructuring, arrow functions, etc. This will not only make development easier, it will make it less error prone and more maintainable and future proof.

I am really interested to update this library and make it easier to use, but I don't want to be bogged down by the requirement of having to use ES5 only unless there are really compelling reasons. 

As I said above, I think for users currently stuck on Node v4, they can get by with the current versions of the library (in conjunction with my sendgrid-mailer package if they want an easier API). But if we are making a new major version, I don't see why we need to limit ourselves.

thebigredgeek:
We could also just use babel with es2015.  The outputted code is mostly readable and we could then target every node version.  I would personally recommend that over cutting support for Node v4.  I know a ton of people who are locked into using it at enterprise companies... and unfortunately those companies are likely paying the most money to Sendgrid.  I refuse to build products these days without v6, so I understand the sentiment, but I think it would be inappropriate to build the formal JS lib for sendgrid in a way that alienates enterprise engineers

adamreisnz:
I think the current one already alienates them if the comments in this thread are anything to go by, so I wouldn't worry about that too much.

thebigredgeek:
Sure, but I think we should be honest about the tradeoffs and fair to the folks at Sendgrid, since they own this library ;)

SendGridDX:
Hello Everyone!

Thanks for the continued feedback. Here is our proposed path forward:

First, I apologize for the delayed response.

I wanted to gather some internal data to help support our path forward. We have found a 10% growth this year for the current SDK (v4.X) utilizing our SendGrid v3 API and a 10% migration from the previous SDK that supports the SendGrid v2 API. There is significant traffic that flows from this SDK daily.

1. We put the current version (v4.X) of this SDK on a new branch for which we update it and the npm package for only patch releases (v4.10.x) for the time period that Node v4 is under maintenance.

2. The new version that @adamreisnz proposed, is to be merged into the master branch as a breaking v5.0.0 version of this SDK once it is deemed stable. Prior to that, we will release testable versions on npm (v5.0.0-betaX.X.X) as a beta for community testing.

3. We communicate this to the community in the following ways:
- Notes at the top of the README for each branch explaining the versioning scenarios.
- Auto detect the Node version in the SDK and provide the appropriate messaging at the command line.

Your feedback is greatly appreciated!

With Best Regards,

Elmer

adamreisnz:
Sounds good to me 👍

adamreisnz:
Alright, I'll carry on with the proposed API as above.

thinkingserious:
Awesome, thanks @adamreisnz!

adamreisnz:
@SendGridDX could you clarify something for me? From looking at the API, it doesn't seem like it's possible to send multiple different emails with a different `from` or `reply_to` values, nor with different `content`.

Am I right in assessing that as such, the extent to which you can send "different" emails to multiple recipients is then limited to having a different subject (through personalizations) and different substitutions only?

If that's the case, then my proposed API for sending multiple different emails to multiple recipients will have to be built such that it sends multiple requests. E.g. we can't do that in a single request.

However, we still need to support the use case of the same email with different subject/substitutions to multiple recipients, which _can_ be done with one request.

I've modified the API a bit to reflect this distinction.

thinkingserious:
@adamreisnz,

You are correct.

1. In our [C# implementation](https://github.com/sendgrid/sendgrid-csharp/blob/master/src/SendGrid/Helpers/Mail/MailHelper.cs#L85), we assume they will use substitutions to differentiate the content. I don't think it will hurt to offer both options (single vs. multiple requests)

2. Could you please elaborate on the usage of the `individual` flag? My first thought is that it should be necessary.

adamreisnz:
@thinkingserious did you mean _shouldn't_ be necessary?

I think it is, at least for the simple API, or there will be no way to differentiate between the two use cases:

1) Send a single email to two recipients who see each other as recipients:

```js
const data = {
  to: ['recipient1@example.org', 'recipient2@example.org'],
  from: 'sender@example.org',
  subject: 'Hello world',
  text: 'Hello plain world!',
  html: '<p>Hello HTML world!</p>',
};
sendgrid.mail.send(data);
```

2) Send two single emails to two distinct recipients, who don't see each other:

```js
const data = {
  individual: true,
  to: ['recipient1@example.org', 'recipient2@example.org'],
  from: 'sender@example.org',
  subject: 'Hello world',
  text: 'Hello plain world!',
  html: '<p>Hello HTML world!</p>',
};
sendgrid.mail.send(data);
```

Example 2 would send two requests to the API, and example 1 would send a single request with two recipients in the personalizations array.

Unless you have another suggestion on how to differentiate these two cases while still keeping the API simple, I think the flag does the job. But I'm open to suggestions, as the use case 1 is not something I had considered before with my wrapper.

thinkingserious:
@adamreisnz,

Thank you for elaborating, that makes sense to me.

My original thought that was for Example 2, you would create your separate calls by looping through your email list, setting one email per data object. Your solution effectively hides that functionality cleanly.

adamreisnz:
Yes, focus is on simplicity, but perhaps it's cleaner to do something like the following instead:

```js
sendgrid.mail.sendMultiple(data);
```

Where we don't use a flag, but instead explicitly call another method to indicate we want to send multiple requests.

thinkingserious:
@adamreisnz,

I like it.

In C# we did:

`CreateSingleEmail`
`CreateSingleEmailToMultipleRecipients`
`CreateMultipleEmailsToMultipleRecipients`

Those are `SendGridMessage` object builder (the result is the request body), which is then passed to `SendEmailAsync`.

[Here is an example](https://github.com/sendgrid/sendgrid-csharp/blob/master/ExampleCoreProject/Example.cs#L23).

Generally, I like the idea of having an EmailMessage object builder (that handles various scenarios), that then gets passed to a method that handles making the API call.

adamreisnz:
@thinkingserious it's ok to have such object builders, but I like to hide them behind the scenes for simplicity, as that's the biggest problem with the node js library at the moment imo. 

These methods would still be exposed publicly, so people can use them if they want, but I'd also like to have the object/config parser API as proposed in this PR for really simple email sending.

I will give it some more thought.

thinkingserious:
Thanks @adamreisnz,

I agree, these things can co-exist. In the end, I think what you propose will just make sense to Node.js developers, and that is the end goal. I trust your judgment on that. We already have customer validation on your design. We can always add helper functions later.

I mention the C# SDK, because it's the first SDK we re-wrote, with a lot of community help. That said, the design used there is of course not a direct match here, but I thought the ideas are worth considering. I hope I am not confusing things with those examples.

thinkingserious:
@adamreisnz hopefully this release will make testing smoother :)

https://github.com/sendgrid/sendgrid-nodejs/releases/tag/v5.1.0

adamreisnz:
Great, that's good, I will check it soon.

adamreisnz:
> `sendgrid.mail.sendMultiple(data);`

I will be going with this option rather than the flag!

adamreisnz:
@thinkingserious I have now refactored the mail endpoint and helpers and will push this code up for review soon.

A couple of comments before I proceed:

I've stripped out some helper classes that in my opinion were adding unnecessary complexity. My thinking is, if someone needs to use the advanced options/settings of the mail endpoint, they will probably be smart enough (or have smart enough JS devs) to construct the required objects themselves, as they would have been looking at the API specifications online. Wrapping every single setting in a helper class just makes it needlessly convoluted and hard to maintain. JS devs are used to working with API's and objects so this shouldn't pose a problem or create a barrier in my opinion.

There are two more helpers in the code base, a contact importer and inbound parser. To me, they feel slightly out of place in this package. I would prefer to see them in a separate node package, so that we can separate concerns and avoid bloat. We can't possibly include an API for every single Sendgrid endpoint in a single package. Well, we *could*, but I think really shouldn't. Separating it out keeps it lightweight, easy to maintain individually, and allow people to grab only what they need.

While I started the rework with the idea that we could add multiple endpoints, I not feel that this is incorrect. I would strongly advocate to develop a package structure as follows:

#### sendgrid-api
This will contain the Sendgrid class that handles creating requests and sending them to the API. It can be used standalone to create requests to endpoints that have no helpers written for them, but it will also be used by all the helper packages.

### sendgrid-mail
This will be dedicated to sending mail, and contain my current re-worked API. It will be a helper on top of the `sendgrid-api` package to facilitate interaction with the mail endpoint.

### sendgrid-contactdb
This will be a dedicated package for dealing with the contactdb endpoint, and contain amongst others the contact importer.

Etc.

All of these packages can alternatively be published under the `@sendgrid` namespace to avoid collisions with possibly existing packages, e.g. `@sendgrid/api`, `@sendgrid/mail`. Or, if there aren't any collisions the above suggested names could be used instead. Personally I would choose for the namespace because it'll be clear this way which packages are officially maintained by Sendgrid and you'll have them all grouped together.

Keen to hear your thoughts on this.

If you want to go ahead with splitting the packages (which I really recommend -- it's very common to do that in node land), then I would suggest we create new packages where I will then put the new code, and leave the current package for legacy purposes.

thinkingserious:
Hi @adamreisnz,

A bit of background. The original concept for this SDK, was to first release a base version that can successfully call all v3 endpoints and decouple from the v2 mail endpoint. I believe this is the equivalent of your suggestion for `sendgrid-api` and is the current state of this SDK.

Additional functionality was to be implemented as helpers, similar to a plugin type of model. We envision that over time there would be many useful helpers covering a variety of use cases. Handling attachments, for example. We started down this route with a very basic mail helper to serve as a model.

That said, we are open to adjusting the helper model to best fit the community needs. Based on your comments, I like the idea of splitting things into a sane package structure as you suggest above.

I have few questions:

1. Assuming we use the package naming you suggested, how would the directory structure look like in this repo (Do they all stay in this repo?)? For tests as well. Or are you suggesting separate repos also.
2. How would deploying to npm work in this scenario? I think you are suggesting that each of the package examples you provided would be deployed as it's own package, where all packages are dependent on the `sendgrid-api` package.

Thank you so much for all the time you have spent on this. I appreciate your thoughtfulness and look forward to your reply :)

With Best Regards,

Elmer

adamreisnz:
Hi @thinkingserious 

Yes I think we are mostly in agreement then. Indeed the `sendgrid-api` that I proposed would be the package that can successfully call all V3 endpoints, in essence the current package.

So I would advocate to remove the helpers from it, and put them in separate packages, each with their separate repository as well.

This means you can update each package separately without it ever affecting another one, plus you can have different collaborators for different packages. It would also separate out tests and documentation, again making it all easier to maintain and develop.

The main api package could of course have references/links to the helper packages so they can easily be found.

Your assessment at 2 is also correct, each package would be deployed to npm separately. I will include a simple script that will make it easy to release a new version to npm. Each package will indeed have the `sendgrid-api` packages listed as a dependency, or maybe better, a peer dependency.

I could set this up on my own github account initially so you can have a look around the code structure and see how it would look. If you're happy with that, we can look at creating the repo's under Sendgrid's account and look into what names to use for publishing the packages on npm.

thinkingserious:
That sounds great @adamreisnz. I think we are in sync. Please do setup the example in your GitHub to help ensure I don't screw it up ;)

I hope a few others in the community will weigh in before we pull the trigger. (hint, hint @thebigredgeek, @IonicaBizau, @fisch0920, @SPARTAN563 to name a few) :)

adamreisnz:
Excellent, I'll get on it!

gastonelhordoy:
Hey @adamreisnz, giving the multi-project structure proposed, I think you might benefit from [lerna](https://lernajs.io/). Give it a try, you can also publish to npm independently

adamreisnz:
@gastonelhordoy thanks, I will look into that.

SPARTAN563:
Sounds good to me, don't think it'll significantly affect the complexity of building TypeScript definitions for it and may, in fact, make them easier for others to help maintain going forward (always a plus).

IonicaBizau:
Sounds good to me too! 😄

adamreisnz:
@thinkingserious so would you want to use https://lernajs.io/ ?
This way we can seemingly keep everything in one repo, but still publish as separate packages.

I haven't worked with it myself, but having managed many similar packages across an organisation, I can see the appeal. On the other hand, having separate repositories does have its benefits as well, as it allows you to separate ownership and collaborators, without everyone getting in each other's way.

@gastonelhordoy can you provide us some more details on how it worked for you? Are there any negative side effects to using it or things we'd need to consider?

thinkingserious:
@adamreisnz,

I'm not opposed to separating out into different repos, like we did with nodejs-http-client. 

I like your follow up question to @gastonelhordoy. In the end we need to weigh the pros/cons of both approaches.

gastonelhordoy:
Hey @adamreisnz, sorry for the late reply.

It worked pretty well for us. Managing separate packages that are part of the same bundle is a bit tedious and error prone in same scenarios, specially if there are breaking changes and you have to update the dependency required by dependants.

For open source projects I think the only benefit having separate repositories is, as you said, the ability to have different contributors and access levels to each module. Having said that... you still have ACL and PRs in a monorepo, so I think it's not a big problem.

Apart from the version management benefit, adopting lerna has also had a positive impact in our development process, since a change in an internal dependency is automatically available (symlink in `node_modules`) in the dependant module, versus the publishing & re-installing flow.

thinkingserious:
Thanks for taking the time to reply @gastonelhordoy!

I really like the idea of minimizing the impact of breaking changes. 

Do you have an open sourced project using lerna that you can share with us? Or do you have an example of a project that is using lerna effectively?

gastonelhordoy:
Hey @thinkingserious, my pleasure! We all will also benefit from a smoother development process for this project.
I was going to paste the link in my previous comment and forgot about it... as a matter of fact Babel uses lerna, [check out their reasoning](https://github.com/babel/babel/blob/7.0/doc/design/monorepo.md). They also include a list of other (BIG) open source projects who have adopted a monorepo approach even though they don't use lerna.

thinkingserious:
Haha: "Juggling a multimodule project over multiple repos is like trying to teach a newborn baby how to ride a bike."

Well, they do make a compelling case. [This ](https://danluu.com/monorepo/) was also an interesting read.

What do you think is the best argument against them? Here is one I found:
- [Ex Googler](https://news.ycombinator.com/item?id=10140635)

adamreisnz:
I guess it can make sense. This way you will be able to keep all individual packages versioned the same version together easily, instead of having them all in separate versions which can be confusing.

Also, the one place to report issues is tempting, although again, with multiple collaborators working on certain different parts, it might make it a bit harder to get the right person to respond to a particular issue. For example, if someone were to manage a specific endpoint package only, and then a bunch of issues were raised related to another endpoint, they might disable watching the repo because I'd get sick of the notification emails. This could result in issues being left unattended for a long time. Poor @thinkingserious would have to keep track for every issue who's the responsible collaborator and mention them so they can get on top of it 😉 

I think the comments from the Ex Googler are not as relevant for this project. It's not going to be *that* huge of a repo and SVN was notoriously slow compared to GIT. Because it's all hooking up to the same API, it would kind of make sense to keep it grouped together. And have the test run together is sensible too.

I suppose we can give it a shot?

thinkingserious:
I'm up for giving this a shot :)

I just wanted to hear from at least one strong opposing viewpoint that would be relevant to this project. From my quick search, it seems most people are onboard the monorepo train and they provide some compelling arguments.

adamreisnz:
Ok, so the question becomes, do we do this in this repo, and if so, how will that work when currently this repo links to one package.

@gastonelhordoy any tips on how to best migrate an existing repo?

Should we just create the structure in a new branch, then wipe the code, initialize with lerna, put the code back and setup the packages accordingly? Or are there any other steps involved?

I feel that maybe it's best to start with a clean repo, so we don't have to worry too much about breaking existing packages and so we don't get stuck with the old "baggage". But that just might be my initial paranoia 😉  

What do you guys think?

thinkingserious:
How about both, use the `@sendgrid` namespace along with using the `sendgrid-` prefix for every package. Or is that frowned upon?

adamreisnz:
I think that's redundant and not necessary, as the package name will in fact have the scope namespace prepended, so you get:

```
@sendgrid/api
@sendgrid/mail
```

Instead of 


```
@sendgrid/sendgrid-api
@sendgrid/sendgrid-mail
```

Angular doesn't use both either, for example: https://www.npmjs.com/~angular
The packages that are scoped, are simply called by their name without prefix.

thinkingserious:
Perfect, that looks nice and clean to me.

gastonelhordoy:
> Ok, so the question becomes, do we do this in this repo, and if so, how will that work when currently this repo links to one package.
> @gastonelhordoy any tips on how to best migrate an existing repo?

>Should we just create the structure in a new branch, then wipe the code, initialize with lerna, put the code back and setup the packages accordingly? Or are there any other steps involved?

> I feel that maybe it's best to start with a clean repo, so we don't have to worry too much about breaking existing packages and so we don't get stuck with the old "baggage". But that just might be my initial paranoia 

I guess there are two alternatives, start from scratch with a clean repo or move things around in this one.
For the clean repo option, you can [keep commits and authoring](https://github.com/lerna/lerna/#import) using `lerna import`. This way you don't run into any major risk and while still having history.

If you go for the second option, I think the best way would be to branch out and move all the existing code in this project to the `packages/core` directory (or `api`). After that you can add the lerna stuff in the rort plus the additional packages. This way you will also keep history without completely wiping out the repo.

What do you think?

thinkingserious:
Starting fresh is always tempting :)

Based on your description @gastonelhordoy, it seems straight forward to move forward with this repo. Therefore, that would be my vote.

gastonelhordoy:
> Based on your description @gastonelhordoy, it seems straight forward to move forward with this repo. Therefore, that would be my vote.

Ya, it also depends on how much time it will take to merge this to master and how much activity (bug fixes) could be applied in the mean time. But staying in this branch will be easier to merge those things and applied potential fixes to the branch.

adamreisnz:
Ok, looks like we'll have to take the current master branch and split that off into a legacy branch then, in order to keep maintaining the current `sendgrid-nodejs` package. But that can be done when we're ready to merge all the mono repo setup into master.

@thinkingserious since this PR is linked to the master branch and github doesn't allow us to move that yet, do you want to create a new `monorepo` branch so that I can create a PR for that branch instead?

thinkingserious:
`monorepo` branch created.

I'm not sure why these tests are failing though, they pass on my local machine. Do that pass for you now @adamreisnz? You just need to run `mocha` and prism will startup and shut down automatically now.

adamreisnz:
@thinkingserious I haven't reached the stage where I run the tests yet, we'll probably have to do some re-work on those anyway once we get the new codebase up.

thinkingserious:
Hmm, well they do work on my local machine. I suppose we can troubleshoot why Travis is angry later.

gastonelhordoy:
@thinkingserious , [here you can see](https://gist.github.com/nolanlawson/457cdb309c9ec5b39f0d420266a9faa4) a more evaluated discussion of why PuchDB drop lerna in favor of a custom layout+toolset. 

Anyways, I think lerna is a pretty good tool and makes sense for a small set of projects like this one. We are using it even for larger projects and we haven't had any major issue for daily use cases.

thinkingserious:
Thanks @gastonelhordoy, that was a good read. OK, I'm convinced we should use lerna for this update :)

adamreisnz:
@thinkingserious ok let's do it!

adamreisnz:
@gastonelhordoy have you run into the issue `Cannot find module 'internal/fs'` when using `lerna bootstrap` before?

I couldn't find a solution for it, so I switched the npm client to `yarn` and that worked fine:

`lerna bootstrap --npm-client=yarn`

gastonelhordoy:
Hey @adamreisnz , I didn't face that issue but I read something similar not specifically using lerna.

Check out [these](https://github.com/nodejs/node/issues/9377) [issues](https://github.com/npm/npm/issues/14232).

Anyways, `yarn` should work just fine as well. You can even set `npmClient` in the `lerna.json` file so you don't have to type it every time.

adamreisnz:
@thinkingserious I've pushed some initial code to the branch in my fork. It shows you the monorepo structure, with currently 5 packages:

### client
This is the client for interacting with the API. I noticed your old implementation used a custom REST client in the form of [sendgrid-rest](https://www.npmjs.com/package/sendgrid-rest). I've dropped this in favour of [http-as-promised](https://www.npmjs.com/package/http-as-promised) which is a promise based wrapper around [request](https://www.npmjs.com/package/request).

### contact-importer
Not sure if this is used much but I've turned it into a separate standalone package, so that people who want to send mail don't need to fetch all the dependencies of the contact importer.

### inbound-mail-parser
Same.

### mail
The mail service, basically the thing we're trying to improve here! :)

### mail-helpers
I've moved some of the classes that are still used into this separate helpers package. This way the other packages (or users) can use them if needed.

---

Could go over the structure and see if everything looks in order to you?

You may notice that I have removed all tests and examples for now, because I wanted to restructure and clean those up as well. I will probably do a few, and leave it up to you or someone else to do the rest if that's ok.

I've also created a new readme for the client and mail service, based on the work we're doing here. You might want to go over those later and add any other information/links from the old readme as needed. You might also want to write a general readme in the root of the monorepo, which I've left empty for now.

I will create some tests and examples now to validate the new API and see how it works.

adamreisnz:
@gastonelhordoy yes I've tried to troubleshoot with those issues, but nothing helped except using yarn. Thanks for the tip, I will add it to the JSON.

adamreisnz:
@thinkingserious I'm actually not that happy about the Prism being integrated with the test suite. This makes it hard to decouple from it, and the long timeout needed to wait for it to start is a real headache if you want to run the tests often.

For that use case, it would be far simpler to have a `npm run` script that starts up prism for you (can check if installed and all that), and just have it run in the background while you're testing.

Is this approach needed to run the tests on Travis? Or could we decouple it and create a separate mechanism to launch prism prior to testing?

thinkingserious:
@adamreisnz,

Nice work on the monorepo structure!

I think we should rename `mail-helpers` to `mail-legacy` if we are replacing the previous mail helpers with the `mail` package. Or, is `mail` dependent on `mail-helpers`?

The reason why we depend on `sendgrid-rest` is so that we [reduce the third party dependencies](https://github.com/sendgrid/nodejs-http-client/blob/master/package.json#L21). I don't think we need to go beyond node's http/https native libraries for our use cases. That said, having the option to pass your own http client would be ideal.

For Prism, I love the idea of being able to manually start and stop it for local testing. The automation was introduced to make contributing and CI easier. I think we can also optimize that 20 seconds of wait time as that was done somewhat arbitrarily. We just need to make sure Prism is running before we run the tests against it.

I will dig deeper to assist soon, unfortunately I have several items ahead of this on our backlog, but I will continue to monitor your feedback and respond ASAP. Thanks!

With Best Regards,

Elmer

adamreisnz:
@thinkingserious 

> I think we should rename mail-helpers to mail-legacy if we are replacing the previous mail helpers with the mail package. Or, is mail dependent on mail-helpers?

They are still dependent on it, but it only contains some of the top level classes, which I've also refactored slightly. I am still playing around with their structure to see if we can make better use of them. Fact of the matter is, the "personalizations" structure in the API is, while flexible, quite complex, so having these helpers internally will assist us navigate that data.

> The reason why we depend on sendgrid-rest is so that we reduce the third party dependencies. I don't think we need to go beyond node's http/https native libraries for our use cases. That said, having the option to pass your own http client would be ideal.

I'm not convinced. Instead of introducing a dependency on a 3rd party package, you introduce a dependency on your own package, which then needs to be separately maintained and kept secure and up-to-date. The `request` package has over 1 million downloads a day, it's being used extensively, and is the de-facto standard for sending requests. The only reason I use the `http-as-promised` wrapper around it, is to be able to use promises. However, we can skip that and I can write my own promise wrapper around `request` if you think that's better.

I would however strongly recommend relying on existing, popular packages, rather than writing our own. It simplifies things so much, like for example we no longer need to detect manually now if the response was valid or not, which was causing trouble in the original implementation of this library.

> For Prism, I love the idea of being able to manually start and stop it for local testing. The automation was introduced to make contributing and CI easier. I think we can also optimize that 20 seconds of wait time as that was done somewhat arbitrarily. We just need to make sure Prism is running before we run the tests against it.

So I think for local testing, we can just use the start/stop scripts to run it. We can enable the automatic running by introducing an environment variable that is enabled on Travis, if you want it to run automatically there.

I noticed the 20 seconds seemed arbitrary, but also saw that starting Prism once took 5 second and another time 10 or so, so it does fluctuate. I will try to come up with a solution for both local testing and CI.

Do you know if Prism needs to be downloaded every time for the CI, or is it part of the pre-configured package?

> I will dig deeper to assist soon, unfortunately I have several items ahead of this on our backlog, but I will continue to monitor your feedback and respond ASAP. Thanks!

No worries, I work on it when I can as well, thanks for monitoring the thread.

thinkingserious:
@thinkingserious 

> I think we should rename mail-helpers to mail-legacy if we are replacing the previous mail helpers with the mail package. Or, is mail dependent on mail-helpers?

>> They are still dependent on it, but it only contains some of the top level classes, which I've also refactored slightly. I am still playing around with their structure to see if we can make better use of them. Fact of the matter is, the "personalizations" structure in the API is, while flexible, quite complex, so having these helpers internally will assist us navigate that data.

>> But for now, they are just an internal dependency that is used by some of the other packages.

Where I'm getting stuck is that `mail` as you have defined, is what I envisioned as the `mail-helper` where `mail-helper` as you have defined is the supporting model for it. I'm not sure what naming to recommend just yet though.

> The reason why we depend on sendgrid-rest is so that we reduce the third party dependencies. I don't think we need to go beyond node's http/https native libraries for our use cases. That said, having the option to pass your own http client would be ideal.

>> I'm not convinced. Instead of introducing a dependency on a 3rd party package, you introduce a dependency on your own package, which then needs to be separately maintained and kept secure and up-to-date. The `request` package has over 1 million downloads a day, it's being used extensively, and is the de-facto standard for sending requests. The only reason I use the `http-as-promised` wrapper around it, is to be able to use promises. However, we can skip that and I can write my own promise wrapper around `request` if you think that's better.

I'm good with bringing that dependency into the `client` package. I think the native http/https client is proven and tested.

>> I would however strongly recommend relying on existing, popular packages, rather than writing our own. It simplifies things so much, like for example we no longer need to detect manually now if the response was valid or not, which was causing trouble in the original implementation of this library.

If this is the only win we get from adding the dependency, I prefer we fix it, maybe thinning the wrapper around http/https. To give some background, we used to get a fair amount of tickets regarding incompatibilities with a chosen HTTP client. In particular, Guzzle in PHP was the biggest offender. When we rewrote the SDKs to support v3 we did a dependency audit, trying to cut all of them if possible.

I am definitely a fan of using proven solutions when it makes sense, such as you have demonstrated, but only if absolutely necessary. For example, in Java, we had to use a third party HTTP client because the Java native client does not support PATCH.

> For Prism, I love the idea of being able to manually start and stop it for local testing. The automation was introduced to make contributing and CI easier. I think we can also optimize that 20 seconds of wait time as that was done somewhat arbitrarily. We just need to make sure Prism is running before we run the tests against it.

>> So I think for local testing, we can just use the start/stop scripts to run it. We can enable the automatic running by introducing an environment variable that is enabled on Travis, if you want it to run automatically there.

That makes sense.

>> I noticed the 20 seconds seemed arbitrary, but also saw that starting Prism once took 5 second and another time 10 or so, so it does fluctuate. I will try to come up with a solution for both local testing and CI.

Awesome!

>> Do you know if Prism needs to be downloaded every time for the CI, or is it part of the pre-configured package? 

Yes. We do that because otherwise we have to include the prism binary in the repo. 

> I will dig deeper to assist soon, unfortunately I have several items ahead of this on our backlog, but I will continue to monitor your feedback and respond ASAP. Thanks!

>> No worries, I work on it when I can as well, thanks for monitoring the thread.

Thank you for your patience and time, they are both much appreciated!

adamreisnz:
> Where I'm getting stuck is that mail as you have defined, is what I envisioned as the mail-helper where mail-helper as you have defined is the supporting model for it. I'm not sure what naming to recommend just yet though.

I was going for "mail service" for the endpoint service, rather than helper. But yes, I agree, I don't like the name "mail-helpers" much either.

As for the http client, you can rest assured that it supports everything perfectly :) It is in fact a wrapper around the native http methods, just has the added convenience of being comfortable to use and easy to configure. I really think we shouldn't be re-inventing and then maintaining the wheel when such good options are available to us. I can't say much about C++ or Java implementations, but in Node, the eco system of open source packages is so extensive and well tested that I can't imagine you'd run into problems with it.

> Yes. We do that because otherwise we have to include the prism binary in the repo.

Ok, thanks for clarifying!

thinkingserious:
> I was going for "mail service" for the endpoint service, rather than helper. But yes, I agree, I don't like the name "mail-helpers" much either.

Agreed, lets think on this further ...

Maybe we just need the `mail` package.

> As for the http client, you can rest assured that it supports everything perfectly :) It is in fact a wrapper around the native http methods, just has the added convenience of being comfortable to use and easy to configure. I really think we shouldn't be re-inventing and then maintaining the wheel when such good options are available to us. I can't say much about C++ or Java implementations, but in Node, the eco system of open source packages is so extensive and well tested that I can't imagine you'd run into problems with it.

*I really think we shouldn't be re-inventing and then maintaining the wheel when such good options are available to us.*

This is exactly why I favor the native http client. 

The `requests` library has 22 dependencies and I'm just concerned that it will cause maintenance issues down the road. That said, I believe you are much more well versed in the ways of the Node.js community, so I am willing to take this path. However, I need to do a bit more research so that I can prove to our internal team that taking on these additional dependencies will we worth it. Any help you can provide on that front would be much appreciated.

thinkingserious:
I'm still torn on this. 

I did some additional research and found that Twilio uses the `request` library, but Stripe and Nexmo use the native `http` library. These are three companies that I believe invest much into their libraries.

Looking through the `http` library vs the `request` library docs, I'm not sure what use cases `request` achieves that `http` \ `https` can not. 

But, like I said, I'm willing to be convinced otherwise.

adamreisnz:
I didn't get back to this in detail yet, but I don't think it's a matter of what can be achieved by both libraries. I'm sure I could write an HTTP client in plain javascript without using *any* libraries, I'm sure it's achievable. The question is more how much work will it be, how much effort will it take, and how complex will it be to maintain it.

With the `request` library it's pretty much plug and play and away we go. With the native `http`, we need to build our own wrapper, plus helper functions and error handling etc. That's brittle in my opinion, and unnecessary. 

I am all for using as few dependencies as possible, but as happened with one of my other npm packages (https://www.npmjs.com/package/replace-in-file), which I wanted to keep dependency free because it was such a simple functionality, I realised that there are some packages out there that are just awesome at what they do and they are used and trusted by millions of users and that there is no point re-inventing the wheel or struggle with copying their functionality.

From looking at the code of `sendgrid-rest`, I felt it's doing exactly that -- trying to replicate the functionality of `request` basically, and creating an unnecessary new package in the Node eco system that will just add unnecessary work for maintainers. It does basically the same thing as `request` (e.g. a wrapper for native `http`), but it doesn't do a great job and has a lot of features missing.

So my initial thought was, ok we can skip this, maybe it was written a long time ago before `request` was around.

In the end, whether we require `sendgrid-rest` as a dependency or `request`, does it really matter? Yes `request` has more dependencies of it's own, but we don't need to worry about those (much).

Sorry if I'm not making sense, it's still early in the morning :)

thinkingserious:
Thanks for the detailed response!

A bit of background of `sendgrid-rest`. Originally, this functionality was included as a part of this SDK. We extracted it into it's own package thinking that others might benefit from a very simple HTTP client that had no dependencies, given that for our own use case, we did not require too much from a HTTP client.

What I'm proposing is to drop the dependency to `sendgrid-rest` and move that logic into this SDK.

> With the native http, we need to build our own wrapper, plus helper functions and error handling etc. That's brittle in my opinion, and unnecessary. The fact that the sendgrid-rest library didn't indicate to the user that a request had failed was a huge miss for example.

I am willing to tackle this part of the project. With your guidance (and the community) I think we can create something that works great for our use case. I know this will add significant development time, but I'm hoping we will save dealing with compatibility/dependency issues in the long run. In our other libraries, the various communities helped do this (specifically, for functionality that was missing) and I'm happy with the reduction of tickets relating to dependency issues.

If we do end up going with `request`, I would hope we could integrate it in a way that would make replacing it with another HTTP client as easy as possible. 

I'm not against including dependencies, but I want to do so with extreme caution.

Good morning :)

adamreisnz:
If you want to refactor it so it only uses `http` then you're welcome to tackle that :)
It'd be great if it could maintain the same API as currently:

https://github.com/adamreisnz/sendgrid-nodejs/blob/monorepo/packages/client/src/client.js

That way it could be replaced with other libraries, like `request`, easily.
One change, it should probably return a promise and not a stream like `request` does.

I still think it's preferable to go with `request` though. Maybe the other participants in this thread can voice their opinion as well so we can discuss further?

thinkingserious:
Yes, I think that whatever HTTP client we ultimately choose, it should be easily replaceable.

It would be fantastic to have some other voices chime in :)

Any thoughts? @thebigredgeek, @IonicaBizau, @fisch0920, @SPARTAN563

remialvado:
Hi guys, remarkable work from what I can see. Thanks a lot for this : I was going to refactor our node mailer service to migrate sendgrid-nodejs from 1.9 to 5.1 but I thinks I will wait a few more days for this new version to come : the 5.x API is really not developer friendly as far as I can see :(
Thanks again !

adamreisnz:
@remialvado thanks, we are still a bit away from moving this to production, as we need to make a decision around what underlying http client to use. But keep an eye on this PR for updates, hopefully we can launch something usable soon!

sendgrid-gists:
@adamreisnz,

How about we go with what you have created, but make `const http` configurable rather than a const, but with this release, we note that only `http-as-promised` is supported?

thinkingserious:
@adamreisnz,

How about we go with what you have created, but make const http configurable rather than a const, but with this release, we note that only http-as-promised is supported?

adamreisnz:
Ok, but I think I'll refactor that to use `request` on account of that being a more popular and trusted service. `http-as-promised` is not used as much, and essentially it's just a wrapper around `request` with a Promise implementation anyway.

Not sure how to make the client configurable, as it will be added as a dependency (the only one) of `sendgrid-client`, which in turn will be used by all the endpoint services to make API calls. If we make it configurable, we will need to add multiple different clients as a dependency, unless you'd want to switch to peer deps instead. But that will complicate it needlessly for the consumers I think. We'd also have to ensure that `sendgrid-client` can handle all the different clients, as they might have different API's.

We might not really be understanding each other here 100%, but to keep this project going, I'll carry on and finish the packages as I think they make the most sense for now and we can then review with the rest of the people involved here in this thread, agreed?

thinkingserious:
Agreed and thanks!

ameesme:
Any news on the HTTP-client decision? I'd really like to implement this new API, rather than having to refactor our code again soon after implementing the "old" one.

adamreisnz:
Decided to go with request for now, I'll try to work on it and have a first
draft ready this week. Sorry progress has been a bit slower than I would
have liked due to other projects absorbing majority of this wonderful asset
called time.

ameesme:
@adamreisnz Of course! Thanks a lot for your work on this.

adamreisnz:
Quick update, have implemented the `request` library and update the mail helpers to support all properties as defined in the V3 API, and handle both camelCase and snake_case property names.

Needs a bit more work on handling attachments which I'll get to tomorrow.

I've refactored the package basically according to the following premises:

- Handle all the cases as described in the original post in this thread.
- Provide an easy to use and understand API for 90+% of the use cases
- Support all the advanced configuration settings as defined in the V3 API for people that need those settings.
- Assume that developers who need the more advanced settings are able to read the API documentation and capable of creating the required data structures themselves.
- Support classes & helpers are primarily for internal use and the API should be able to send email without forcing you to use them.

Will take it for a spin tomorrow as well to see how it holds up. I'll need some help from other members to flesh out the new test suite and create typings etc.

thinkingserious:
That's awesome @adamreisnz!

Just FYI and perhaps this is helpful for this SDK, but we are starting to initiate this project across all the SDKs, if you have a moment, please check out the initial ideas (particularly the kitchen sink) for [PHP](https://github.com/sendgrid/sendgrid-php/issues/413), [Ruby](https://github.com/sendgrid/sendgrid-ruby/issues/169) and [Python](https://github.com/sendgrid/sendgrid-python/issues/323). We also updated the documentation for the only SDK where we have completed this work ([C#](https://github.com/sendgrid/sendgrid-csharp/blob/master/USE_CASES.md)). 

Thanks!

adamreisnz:
I've created a new PR which hooks into the actual `monorepo` branch: https://github.com/sendgrid/sendgrid-nodejs/pull/406

Feel free to examine the code and structure there.

adamreisnz:
@thinkingserious I've updated the readme to cover most aspects of the new API, and also created units tests for all the helpers.

I'd like to create some more unit tests that actually use the prism API but I'm not sure if I'm using it correctly. Could you have a look at the `packages/mail/src/mail.spec.js` file and point me in the right direction? Not sure how you guys mock your API.

Lastly, would you be able to publish these packages to npm under the new namespace as an alpha or beta version? I assume you'd want to start at the major version `6.0.0` since that follows the `5.0.0` for your current package, however, since technically these packages are going to be different npm packages under a new namespace, you could also just start them all from `1.0.0`. Will leave that up to you to decide on.

Once they are published with alpha or beta suffix I can take them for a spin in my projects. That's not doable now on account of it being a monorepo and the references to the packages would break when I simply pointed to the git repo branch in `package.json`.

adamreisnz:
Hi @thinkingserious, any chance you could have a look at publishing those packages sometime this week? I'll have some more time this week to wrap this up and test it, am worried that if we leave it for too long another month will pass ;)

thinkingserious:
HI @adamreisnz,

I will definitely get to it early this week. Thanks!

adamreisnz:
Awesome thanks!



[image: photo]
Adam Reis
Co-founder at  Reis Ltd
M 022 100 1067 <+64+22+100+1067> E adam@reis.nz W www.reis.nz
<https://www.linkedin.com/in/adamreisnz> <http://twitter.com/adamreisnz>

thinkingserious:
@adamreisnz,

I've published the packages, but they are named a bit differently, see here: https://www.npmjs.com/~thinkingserious

I tried using the @sendgrid/mail convention, but I could not get it to work. I created a @sendgrid organization and tried to upload as scoped packages, but no luck. I kept getting errors about needing to pay for private repos. Then, I looked at the examples on Lerna's home page and it seems (looking at a random sampling) most are going with the naming convention I ended up using.

Also, seems like this file should be deleted now: https://github.com/sendgrid/sendgrid-nodejs/blob/monorepo/package.json

I've merged your branch into the monorepo branch here and have now updated with my changes.

@gastonelhordoy I like the idea of scoped packages, but it seems like this is not a normal convention for lerna based monorepos. Any thoughts?

adamreisnz:
@thinkingserious thanks for publishing. Have you tried following this guide?

https://www.npmjs.com/docs/orgs/publishing-an-org-scoped-package.html

I would have imagined that Sendgrid has (or could get) a paid organisation account for npm if needed.

I don't mind the naming convention with dashes, but my worry is that it might clash with other 3rd party packages and doesn't clearly make the distinction between 3rd party and "official" Sendgrid packages.

Would be good if @gastonelhordoy can chime in and help out.

thinkingserious:
Yeah, I tried that and also this: https://docs.npmjs.com/misc/scope

Turns out the error relating to the payment was related to private repos, which we don't need.

thinkingserious:
Maybe this line is what tripped things up: https://github.com/sendgrid/sendgrid-nodejs/blob/monorepo/package.json#L4

thinkingserious:
I'm going to give it a try again now.

gastonelhordoy:
Hey @thinkingserious, I guess you are talking about [this issue](https://github.com/lerna/lerna/issues/178), is that correct?

thinkingserious:
Maybe :D -- let me try

Thanks!

adamreisnz:
@thinkingserious yeah that shouldn't be private in that package.json

thinkingserious:
@adamreisnz, @gastonelhordoy,

Looks like it worked: https://www.npmjs.com/org/sendgrid

w00t!

adamreisnz:
Nice, awesome!
I will take them for a spin later today when I'm back home. If you find any issues in the meantime let me know in the Monorepo PR ( #406 ) to keep the noise down in this thread.

ameesme:
Wow, that went through a lot quicker than I expected. The new package looks great! I'll be implementing it in the coming days. Congratulations on the release and thanks for the hard work, especially @adamreisnz! 👍 💪

adamreisnz:
@ameesme thank you! Just bear in mind it's probably still rough around the edges and I haven't tested it extensively yet. Probably should have released it as an alpha version first :)

I hope to test it a bit myself today and tomorrow. 

So probably don't use it in production _just yet_, but keen to hear your feedback! 😄

jamesdixon:
If @adamreisnz isn't on the SendGrid payroll, the man should get free SendGrid fo life!

Nice work, sir.

adamreisnz:
I've hooked up the new packages in one of our apps on the test environment, and it seems like it's working ok! I haven't tested all mail endpoint API properties yet, but I am fairly confident that the library can be tried out by the wider public now. Feel free to give it a try via:

````
yarn add @sendgrid/mail@6.0.0-beta.2
```

When prompted, choose the 6.0.0-beta.2 version for the helper package as well.

I will keep testing and fixing as needed.

thinkingserious:
Hi @adamreisnz,

Your setup instructions worked perfectly!

I tried a few test cases and getting started is super easy :)

Do you have a list of what needs to be completed? For example, it appears that cc and bcc is not implemented.

Thanks!

With Best Regards,

Elmer

adamreisnz:
@thinkingserious I just noticed as well that I had omitted the cc/bcc implementation as per the API specs. I have now included this, but it needs a bit more work to also allow only using the cc/bcc fields for example. I will try to finish this up tomorrow.

All the rest of the settings should work!

thinkingserious:
Thanks!

The kitchen sink example [here](https://github.com/sendgrid/sendgrid-ruby/issues/169) should be a good sanity check. I will need to create the equivalent of this documentation in our USE_CASES.md file.

thinkingserious:
@adamreisnz,

Changes published as version 6.0.0-beta.3. BTW, you don't need to change the version number when you update, Lerna does this for us :)

adamreisnz:
Great! 
I'll implement a kitchen sink example with all settings used and add it to the readme.

adamreisnz:
@thinkingserious I have now also deprecated https://www.npmjs.com/package/sendgrid-mailer so if that's still on the list of 3rd party packages feel free to remove it as it won't be needed anymore.

thinkingserious:
Thanks for the heads up, I've put in a PR to do so: https://github.com/sendgrid/docs/pull/2788

adamreisnz:
Quick update, I've improved a couple of things regarding recipient handling and error response handling, as well as updated the readme with some additional examples and use cases.

There's a few things still not 100% around the handling of cc/bcc's. I'm working on fixing that now and adding some tests for the helper classes to make sure they convert to the proper JSON.

ameesme:
Just implemented 6.0.0-beta.3 and it works beautifully!

adamreisnz:
Awesome, please update to beta 4 for some fixes around cc/bcc handling!

johndatserakis:
Been following this thread for a while now. Thank you @adamreisnz, @thinkingserious, and others for their work on this.

Very excited to use the new progress - but I'm having a strange issue installing the package. Just to warn - there's a very good chance I'm just being incredibly dull.

When I go to `npm install @sendgrid/mail` I get this error: 

`404 Not Found: @sendgrid/mail-helpers@^6.0.0-beta.0`

And I get a little bit more of an error when manually adding it to my `package.json` and `npm install` 'ing from there:

`No matching version found for @sendgrid/helpers@^6.0.0`
`In most cases you or one of your dependencies are requesting a package version that doesn't exist.`
`It was specified as a dependency of '@sendgrid/client'`

Perhaps it's just a small tweak I need when installing? Thank you.

adamreisnz:
@johndatserakis indeed it is. The first version published for this package had the beta suffix, and npm doesn't seem to handle that well when running just `npm install @sendgrid/mail`. You can fix it by explicitly specifying the latest version:

`npm install @sendgrid/mail@6.0.0-beta.4`

or

`yarn add @sendgrid/mail@6.0.0-beta.4`

If/when prompted, also choose the beta 4 version for the helpers package.

It should work as expected again once the "full" version is released.

johndatserakis:
Awesome thanks @adamreisnz - when using yarn the option to choose the helper version came up fine - although it didn't offer any such option in npm. It's nice that yarn does that, hopefully npm will adopt that (it might already and I'm just doing it wrong). Although as you mentioned in the future it won't be needed. Thanks again.

adamreisnz:
No worries, I think we can release a first version soon.

adamreisnz:
How's everyone that has tested the new API doing with it? Encountered any issues? Have any additional suggestions?

johndatserakis:
Working well in my usage. 

```
import sgMail from '@sendgrid/mail';
sgMail.setApiKey(process.env.SENDGRID_API_KEY);
...

let email = await fse.readFile('./src/email/welcome.html', 'utf8');
  const emailData = {
    to: ctx.request.body.email,
    from: process.env.APP_EMAIL,
    subject: 'Welcome To Koa-Vue-Notes-Api',
    html: email,
    categories: ['koa-vue-notes-api-new-user'],
    substitutions: {
      appName: process.env.APP_NAME,
      appEmail: process.env.APP_EMAIL,
    },
};

//Initial function call has the async flag. Works great.
await sgMail.send(emailData);
```

I'm loading an html email template using fs-extra, making substitutions, and adding a category. All good - although I noticed I couldn't use a $ sign in my potential substitution target in my html. No biggie though - just a simple pivot.

Great work. I also really like how simple constructing and sending an email was.

thinkingserious:
Thanks for taking the time to provide feedback @johndatserakis!

adamreisnz:
Thanks for the feedback! @thinkingserious is the $ substitution issue related to Sendgrid's templating engine you think?

thinkingserious:
@adamreisnz,

I think so, I've seen similar issues in the past. Fortunately, it looks like that system is getting an overhaul shortly. Please bring up that issue when you chat with Matt :)

With Best Regards,

Elmer

danteata:
Hi @adamreisnz,
Any chance using this on a CI platform at the moment?

adamreisnz:
@danteata I think the dependency for one of the packages was invalid and manually set to `^6.0.0`, which didn't match the current beta version, and Lerna didn't seem to have updated it.

@thinkingserious I have pushed a change and updated the version, could you release a new beta and check if the versions in all the packages are correct? Then the install should go without hiccups.

thinkingserious:
Thanks @adamreisnz! I should be able to dig in within the next few days. My apologies for the delay.

haydenbleasel:
Hey @adamreisnz, amazing work with this so far. Followed your initial instructions for `Single email to one recipient` with the following data:

```
{
    to: 'xxx@gmail.com',
    from: {
        email: 'xxx@gmail.com',
        name: 'xxx'
    },
    subject: 'hello',
    html: '<p>hi</p>',
    customArgs: {
        foo: 'bar'
    },
}
```

But it results in the following error:

```
Failed to send email Error: Object expected for `substitutions`
    at Personalization.reverseMergeSubstitutions (.../node_modules/@sendgrid/helpers/classes/personalization.js:251:13)
```

adamreisnz:
Hi @haydenbleasel, thank you!

Yes, I fixed that bug but still waiting on @thinkingserious to release a new version. You can use beta.3 for now which doesn't have this bug.

thinkingserious:
My apologies for the delay @adamreisnz & @haydenbleasel. I will be posting here when I get the chance to update to the new version.

adamreisnz:
6.0.0-beta.5 has now been released, which has fixed the issue with npm/yarn complaining about invalid dependency. Thanks @thinkingserious !

fisch0920:
why doesn't the default `@sendgrid/mail` 404 and not point to the latest?

you should never have your users needing to follow a random github thread in order to figure out why your default package doesn't install properly.

adamreisnz:
@fisch0920 version 6 has not been released yet officially, as far as I know, so the beta version is the last currently available one.

The default package should be the old version still, called `sendgrid-nodejs`.

Once this version gets released as 6.0.0 it will become the default. The docs for this version have already been improved and should be up to date.

@thinkingserious do you have a timeline for the release of 6.0.0?

thinkingserious:
@adamreisnz, @fisch0920,

I don't think it will take longer than early this week to officially launch v6.0.0. If you [subscribe](https://dx.sendgrid.com/newsletter/nodejs) to this repo's release notes, you will get emailed when the release is official.

Thanks!

With Best Regards,

Elmer

thinkingserious:
Hello Everyone!

[v6.0.0](https://github.com/sendgrid/sendgrid-nodejs/releases/tag/v6.0.0) has been officially released! Enjoy!

With Best Regards,

Elmer

fisch0920:
Thanks, everyone!  This is a huge win for the Node Sendgrid community 👍 

Awesome work, @adamreisnz @thinkingserious and everyone else who contributed.

adamreisnz:
Awesome thanks everyone for their feedback and contributions! 🎉

PR #384, opened 2017-04-18
Restructure README example code

sigv:
The proposed change:

- ~switches from `var` to `let` for cleaner scoping of variables.~
- ~switches all function expressions to arrow function expressions as to provide an easier overview.~
- removes trailing commas - although valid and arguably wildly used, not really a part of any standardization.
- adds semicolons where missing - mixing use of semicolons and avoiding them in the same codebase is not a good idea and since most of the examples do have semicolons after statements I am adding them where they were missing instead of removing the ones that were there but could be dropped.
- ensures comments start with a single space for consistency and readability.
- adds an error handling if clause to the first example.
- cleans up the source for displaying the logo as well as the _use cases_ header.

All of the changes, with the exception of the last bullet point, are for maintaining a consistent code style that people are welcome by when they visit the README. No actual changes in how the calls are made to the API/library are being introduced here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,21 +65,21 @@ The following is the minimum needed code to send an email with the [/mail/send H
 ### With Mail Helper Class
 
 ```javascript
-let helper = require('sendgrid').mail;
-let fromEmail = new helper.Email('test@example.com');
-let toEmail = new helper.Email('test@example.com');
-let subject = 'Hello World from the SendGrid Node.js Library!';
-let content = new helper.Content('text/plain', 'Hello, Email!');
-let mail = new helper.Mail(fromEmail, subject, toEmail, content);
+var helper = require('sendgrid').mail;
+var fromEmail = new helper.Email('test@example.com');
+var toEmail = new helper.Email('test@example.com');
+var subject = 'Hello World from the SendGrid Node.js Library!';
+var content = new helper.Content('text/plain', 'Hello, Email!');
+var mail = new helper.Mail(fromEmail, subject, toEmail, content);
 
-let sg = require('sendgrid')(process.env.SENDGRID_API_KEY);
-let request = sg.emptyRequest({
+var sg = require('sendgrid')(process.env.SENDGRID_API_KEY);
+var request = sg.emptyRequest({
   method: 'POST',
   path: '/v3/mail/send',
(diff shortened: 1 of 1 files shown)


thinkingserious:
Thanks @sigv,

I think we need to wait on the change from `var` to `let` (and any other ECMAScript 6 specific features) until this PR is merged: https://github.com/sendgrid/sendgrid-nodejs/pull/378.

sigv:
The change being made is not for the library's codebase - it is for the
example code provided in the README and even more so it already partially
uses arrow functions for promise callbacks on the `master` branch. I would
argue experienced developers will swap them out where appropriate anyway
and beginners can benefit from the shorter codebase and cleaner scoping
provided by `let` without thinking about it, so no harm for anyone - only
benefits.

Do you want me to remove the arrow functions and `let` from this PR so that
it can move forward?

thinkingserious:
@sigv,

The plan is to have this current master branch moved to a new branch to support those who need to use Node v4, and then #378 will become the master.

So for this branch, I'd like to keep it consistent (so thanks for the updates :)), but this version would be for those stuck supporting Node v4, so the examples here should be compatible Node v4.

@adamreisnz,

Do you have any thoughts on this?

adamreisnz:
@thinkingserious Once I'm done with the code restructuring, we will likely need to update and clean up many if not all examples in the Readme files. If @sigv is willing to help out with it that'd be great.

sigv:
@thinkingserious Fair point - if a release is supported, then example code can be copy-pasted for that release as well. Pushed a commit that drops `let` and arrow functions for that.

@adamreisnz I can probably pitch in - just let me know when you are done with the restructure on your part.

sigv:
Is there a plan to merge this into `master` now that `let` keyword is swapped back out with `var` and the arrow functions are replaced with full function expressions? 🤔

thinkingserious:
Yes, it will come back up to the top of our queue shortly. Thanks!

thinkingserious:
Hello @sigv,<br /><br/>Thanks again for the PR!<br />                                              <br />We appreciate your contribution and look forward to                                              continued collaboration. Thanks!                                             <br /><br />Team SendGrid DX

PR #407, opened 2017-07-19
Monorepo

adamreisnz:
@thinkingserious PR from monorepo to master


adamreisnz:
@thinkingserious shall we rename the `mail-helpers` to simply `helpers`? E.g. `@sendgrid/helpers`. That way it will be more general and not mail specific, yet still clarify it's helpers.

thinkingserious:
Hi @adamreisnz,

Looking to the future, I imagine we will have helpers such as attachments, whitelabeling, etc.

Perhaps we adopt this naming convention:

@sendgrid/helper-mail (probably the attachments helper goes here to)
@sendgrid/helper-whitelabeling
@sendgrid/helper-apikeys
etc ...

Then we can have @sendgrid/support or @sendgrid/utilities for supporting functions.

What do you think?

adamreisnz:
My reasoning is this;

One package of helpers that are used internally is easier to maintain than a plethora of different helper packages for each endpoint. Furthermore, having a separate mail-helpers package which is just for the mail package kind of defeats the purpose of having a separate helper package. Then we might as well include those helpers in the mail package itself and not have a separate package.

So the whole idea behind one generic helper package is that we can put commonly used classes, helper functions and other reusable components in here, which can be used by all the other future endpoint packages as well.

I think separating them will needlessly complicate things and integrating them with the endpoint packages will make it less reusable.

I don't think there would be separate helpers for attachments and whitelabeling tbh. Those are properties of the mail endpoint, so we could make a "helper" API for it in the mail service package. It may or may not use some helper classes, but those could easily live in the shared helpers library. That will also make it easier for people to import classes from it, because it would all come from one place and they wouldn't have to import from X different helper packages.

adamreisnz:
@thinkingserious if you're ok with the `helpers` name, could you publish the packages again? Due to the inadvertedly renamed support package to mail-helpers, the references to the support package didn't work.

thinkingserious:
Done.

I can no longer unpublish mail-helpers without contacting npm support. I'm doing that now.

adamreisnz:
Thank you, I'll take them for a spin now

adamreisnz:
@thinkingserious it doesn't seem like @sendgrid/mail was updated, it's still on version beta.0 and it still imported the mail-helpers package somehow

adamreisnz:
Nevermind looks like that was my error

adamreisnz:
Just sent the first successful email from a real app using the new libs, seems to work like a charm :)
I will play around more with all the various fields the API supports.

adamreisnz:
@thinkingserious I've added a bunch of tests and fixed the issue pertaining to cc/bcc's, could you release a new version when you have time?

thinkingserious:
I've published as 6.0.0-beta.4

adamreisnz:
@thinkingserious How do you feel about the readme documentation now? I've tried to include usage instructions for all available API settings, as well as a kitchen sink example. Do you feel comfortable releasing this version as a 6.0.0 and have people start using it, or did you want to look into adding some more tests with prism first?

thinkingserious:
This is awesome news @adamreisnz! I will review the docs and tests and push any changes to your monorepo branch.

adamreisnz:
Excellent. I would like to have more tests with Prism and actual API calls, but as I mentioned in the other thread, I wasn't sure how exactly it works so was keen to get your help there.

adamreisnz:
@thinkingserious Could you release a new version with my latest patch please?

And when shall we have a chat about using Prism?

thinkingserious:
I've published as 6.0.0-beta.5

thinkingserious:
Hi @adamreisnz,

I'm now working on getting this over the finish line. Have you been able to get the tests running yet?

I tried, but when I execute `yarn test:mail` I get `Error: Cannot find module '@sendgrid/client'`

While I wait for your response, I am working on updating documentation on your monorepo branch. I will push my changes there.

adamreisnz:
Good to have you back :)

No, I was waiting for you to get going with the tests. You may want to run the `lerna bootstrap` command before you run that test script, it looks like it hasn't linked the dependencies properly.

thinkingserious:
@adamreisnz,

`lerna bootstrap` did the trick :)

I'm powering through the documentation right now and then I'll get back to the tests. Thanks!

thinkingserious:
@adamreisnz 

I can't seem to get the following to work:

```js
const client = require('@sendgrid/client');
client.setApiKey(process.env.SENDGRID_API_KEY);
const data = [
  "127.0.0.1"
];
const request = {
  method: 'PUT',
  body: data,
  url: '/v3/subusers/test_subuser_name/ips'
};
client.request(request)
.then(([response, body]) => {
    console.log(response.statusCode);
    console.log(response.body);
})
```
because the payload that gets sent is:

```js
{
  "0": "127.0.0.1"
}
```

Here is how it worked before: https://github.com/sendgrid/sendgrid-nodejs/blob/master/test/test.js#L4137

Any ideas?

adamreisnz:
@thinkingserious There was an issue with the merge data helper, I've fixed this in 3be7fd8

thinkingserious:
I'm not quite sure why those final two tests are failing. Locally, they work fine. I'll dig deeper.

adamreisnz:
Awesome thanks for the release! Look forward to ongoing developments of this package.
Did you [deprecate](https://docs.npmjs.com/cli/deprecate) the [old package](https://www.npmjs.com/package/sendgrid) on npm?

thinkingserious:
@adamreisnz,

I did not, but now I have :) 

Thanks for the heads up!

thinkingserious:
Hello @adamreisnz,<br /><br/>Thanks again for the PR!<br />                                              <br />We appreciate your contribution and look forward to                                              continued collaboration. Thanks!                                             <br /><br />Team SendGrid DX

adamreisnz:
You're welcome! Glad I could help out.

PR #408, opened 2017-07-19
Updated sendgrid-rest dependancy version to 2.4.0

apat183:


SendGridDX:
[CLA assistant check](https://cla.sendgrid.com/sendgrid/sendgrid-nodejs?pullRequest=408) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.sendgrid.com/sendgrid/sendgrid-nodejs?pullRequest=408) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.sendgrid.com/check/sendgrid/sendgrid-nodejs?pullRequest=408) it.</sub>

thinkingserious:
HI @apat183,

Please add a unit test [here](https://github.com/sendgrid/sendgrid-nodejs/blob/master/test/test.js#L43) and we should be good to go :) 

I'll then bump the version and add you to the CHANGELOG and create a new release.

Thanks!

With Best Regards,

Elmer

apat183:
Hi @thinkingserious,

I'm at struggling to do the unit test to test new functionality because we actually set the port on all requests in our tests already.

```
  var request = sg.emptyRequest()
  if (TEST_HOST == 'localhost') {
    request.test = true
    request.port = 4010
  }
```
and the reason it works is because in the sendgrid-rest lib we set the port if test is true;
```
if ( endpointRequest.test == true ) {
  var http_request = http
  request.port = endpointRequest.port
} else {
  var http_request = https
}
```
So what I could do is remove the line `request.port = endpointRequest.port` from the sendgrid-rest since it's currently redundant based on the updated change. And maybe update the doc examples in sendgrid-nodejs to include an example of setting port?

Let me know your thoughts?

thinkingserious:
Good catch @apat183, I think your approach makes sense.

adamreisnz:
@thinkingserious this can be closed as we no longer use the sendgrid-rest dependency

PR #425, opened 2017-08-17
Export all type definitions in index.d.ts

pvatterott:
It currently isn't possible to use SendGrid's types in a Typescript application as only the constructor is exported. For example, the following code does not work:

```typescript
import * as SendGrid from 'sendgrid';

const email: SendGrid.mail.Email = new SendGrid.mail.Email('test@test.com');
```

The type definition on `email` causes the typescript error: `[ts] Cannot find namespace 'SendGrid'.`

Exporting all of the namespaces allows users to explicitly declare types for variables using SendGrid's type definitions. Explicitly calling `export default` maintains compatibility while also giving access to SendGrids namespaces.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,10 +1,11 @@
-declare namespace SendGrid.Helpers.Error {
+export declare namespace SendGrid.Helpers.Error {
     export interface SendGridError extends Error {
 
     }
 }
 
-declare namespace SendGrid.Helpers.Mail {
+
+export declare namespace SendGrid.Helpers.Mail {
     export interface Helper {
         Email: typeof Email;
         Mail: typeof Mail;
@@ -433,7 +434,7 @@ declare namespace SendGrid.Helpers.Mail {
     }
 }
 
-declare namespace SendGrid.Rest {
+export declare namespace SendGrid.Rest {
     export const emptyRequest: Request;
 
     interface Request {
@@ -458,7 +459,7 @@ declare namespace SendGrid.Rest {
     }
 }
 
-declare namespace SendGrid {
+export declare namespace SendGrid {
     export interface SendGridConstructor {
         (apiKey: string, host?: string, globalHeaders?: { [header: string]: string; }): SendGrid;
         constructor(apiKey: string, host?: string, globalHeaders?: { [header: string]: string; }): SendGrid;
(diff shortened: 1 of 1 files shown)


SendGridDX:
[CLA assistant check](https://cla.sendgrid.com/sendgrid/sendgrid-nodejs?pullRequest=425) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.sendgrid.com/sendgrid/sendgrid-nodejs?pullRequest=425) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.sendgrid.com/check/sendgrid/sendgrid-nodejs?pullRequest=425) it.</sub>

thinkingserious:
Thanks @pvatterott!

@SPARTAN563 thoughts?

SPARTAN563:
Idea sounds good, I'd just like to check the code against some example projects to make sure there are no hidden gotchas with the use of `default` (which doesn't behave exactly the same way as `export =`. I'll let you know when I've confirmed that works correctly.

adamreisnz:
@thinkingserious this probably needs to be adjusted to the new version (6.0.0)

thinkingserious:
@adamreisnz,

This was just [released today](https://github.com/sendgrid/sendgrid-nodejs/releases/tag/v6.1.0), courtesy of the amazing @SPARTAN563.

larssn:
Good work guys but I'm getting this on 6.1.1, in Typescript:

`[ts] Module '"./node_modules/@sendgrid/mail/index"' resolves to a non-module entity and cannot be imported using this construct.`

Thats using: `import * as sgMail from '@sendgrid/mail';`

SPARTAN563:
Hi @larssn, that's a result of the nature of the new SendGrid library's API. You should instead be importing using `import sgMail = require('@sendgrid/mail');` in your code.

There's a brilliant StackOverflow answer which explains why this is the case: https://stackoverflow.com/a/39415662/1098797

Please let me know if you run into any other issues.

larssn:
@SPARTAN563 Thanks Benjamin for the quick reply. Will do.

thinkingserious:
@pvatterott,

Could you please [fill out this form](https://dx.sendgrid.com/swag) so that we can send you some swag? Thanks!

PR #440, opened 2017-09-05
Add default qsStringifyOptions to stringify Arrays correctly

mp-ffx:
A request to `/categories/stats` can contain multiple categories, when supplied as a simple Array, it get transformed to `categories[0]=category1&categories[1]=category2` which get's rejected by the sendgrid API.

When setting the arrayFormat to 'repeat', which transforms to `categories=category1&categories=category2`, everything works as expected.

### DIFF
--- a/packages/client/USAGE.md
+++ b/packages/client/USAGE.md
@@ -82,10 +82,10 @@ For more information, please see our [User Guide](http://sendgrid.com/docs/User_
   "ips": [
     {
       "ip": "192.168.1.1"
-    }, 
+    },
     {
       "ip": "192.*.*.*"
-    }, 
+    },
     {
       "ip": "192.168.1.3/32"
     }
@@ -116,8 +116,8 @@ For more information, please see our [User Guide](http://sendgrid.com/docs/User_
 ```javascript
   const data = {
   "ids": [
-    1, 
-    2, 
+    1,
+    2,
     3
   ]
 };
@@ -215,8 +215,8 @@ For more information about alerts, please see our [User Guide](https://sendgrid.
 
 ```javascript
   const data = {
-  "email_to": "example@example.com", 
-  "frequency": "daily", 
+  "email_to": "example@example.com",
+  "frequency": "daily",
   "type": "stats_notification"
 };
   request.body = data;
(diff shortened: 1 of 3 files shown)


SendGridDX:
[CLA assistant check](https://cla.sendgrid.com/sendgrid/sendgrid-nodejs?pullRequest=440) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.sendgrid.com/sendgrid/sendgrid-nodejs?pullRequest=440) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.sendgrid.com/check/sendgrid/sendgrid-nodejs?pullRequest=440) it.</sub>

mp-ffx:
I've rebased this pull request for easier merging.

mbernier:
@adamreisnz can I please bug you one more time for a review? This is outside my expertise...

adamreisnz:
I'm somewhat reluctant to implement this, for fear it will open a can of worms and we'll be setting defaults for every 50+ of request's [settings](https://www.npmjs.com/package/request#requestoptions-callback).

We already have an API to set default request settings, and I don't see why you can't use that to set the desired query string options, e.g.:

```js
const client = require('@sendgrid/client');
client.setDefaultRequest('qsStringifyOptions', {arrayFormat: 'repeat'});
```

This has the same intended effect, doesn't it?

Request already has sensible defaults, and we provide an option to override any setting, so I don't think it's necessary to explicitly specify additional defaults in this library unless we want to override the defaults that Request is setting.

mbernier:
@adamreisnz I am all for code simplicity. Having to implement and then remember to implement defaults everywhere for future changes is not something I would like to support or ask anyone else to help us support.

I like your suggestion and I'm going to close this PR. Thanks @mp-ffx for suggesting and @adamreisnz for your feedback on this one!

adamreisnz:
No worries -- @mp-ffx if I misunderstood and there are some problems with my suggested approach please let us know.

mp-ffx:
@adamreisnz sounds good, I haven't thought about just using the `setDefaultRequest` method.

I added this to the USAGE.md (#585).

PR #505, opened 2017-10-18
Attachment helper for base64 encoding

spelcaster:
- Added the method readFile to read a file and convert its content to
  base64;
- It's now possible create an Attachment with a raw file;

### DIFF
--- a/packages/helpers/classes/attachment.js
+++ b/packages/helpers/classes/attachment.js
@@ -6,6 +6,8 @@
 const toCamelCase = require('../helpers/to-camel-case');
 const toSnakeCase = require('../helpers/to-snake-case');
 const deepClone = require('../helpers/deep-clone');
+const fs = require('fs');
+const path = require('path');
 
 /**
  * Attachment class
@@ -39,14 +41,34 @@ class Attachment {
     data = toCamelCase(data);
 
     //Extract properties from data
-    const {content, filename, type, disposition, contentId} = data;
+    const {
+      content,
+      filename,
+      type,
+      disposition,
+      contentId,
+      filePath
+    } = data;
+
+    if ((typeof content !== 'undefined') && (typeof filePath !== 'undefined')) {
+      throw new Error(
+        `The props 'content' and 'filePath' cannot be used together.`
+      );
+    }
 
     //Set data
-    this.setContent(content);
     this.setFilename(filename);
     this.setType(type);
     this.setDisposition(disposition);
     this.setContentId(contentId);
+    this.setContent(filePath ? this.readFile(filePath) : content);
+  }
+
+  /**
+   * Read a file and return its content as base64
+   */
+  readFile(filePath) {
+    return fs.readFileSync(path.resolve(filePath));
   }
 
   /**
@@ -54,15 +76,32 @@ class Attachment {
    */
   setContent(content) {
     //Duck type check toString on content if it's a Buffer as that's the method that will be called.
-    if (typeof content === 'string') {      
+    if (typeof content === 'string') {
       this.content = content;
-    } else if (content instanceof Buffer && content.toString !== undefined) {      
+      return;
+    } else if (content instanceof Buffer && content.toString !== undefined) {
       this.content = content.toString();
-    } else {
-      throw new Error('`content` expected to be either Buffer or string');
+
+      if (this.disposition == 'attachment') {
+        this.content = content.toString('base64');
+      }
+
+      return;
     }
 
-    return;    
+    throw new Error('`content` expected to be either Buffer or string');
+  }
+
+  /**
+   * Set content
+   */
+  setFileContent(content) {
+    if (content instanceof Buffer && content.toString !== undefined) {
+      this.content = content.toString('base64');
+      return;
+    }
+
+    throw new Error('`content` expected to be Buffer');
   }
 
   /**
--- a/packages/helpers/classes/attachment.spec.js
+++ b/packages/helpers/classes/attachment.spec.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const fs = require('fs');
+const path = require('path');
 
 /**
  * Dependencies
@@ -18,7 +19,7 @@ describe('Attachment', function() {
 
   //Set content as string
   describe('setContent(), string', function() {
-    it('should set string as content', function() {      
+    it('should set string as content', function() {
       attachment.setContent("Just a string.");
       expect(attachment.content).to.equal('Just a string.');
     });
@@ -26,7 +27,7 @@ describe('Attachment', function() {
 
   //Set content as stream
   describe('setContent(), stream', function() {
-    it('should convert stream to string and set as content', function() {      
+    it('should convert stream to string and set as content', function() {
       const fileData = fs.readFileSync('./packages/helpers/attachment.txt');
       attachment.setContent(fileData);
       expect(attachment.content).to.equal('Just a little file for testing attachments.');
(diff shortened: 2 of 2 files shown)


spelcaster:
```js
//sample file
const sgMail = require('@sendgrid/mail');

const {
  classes: {
    Attachment
  }
} = require('@sendgrid/helpers');

let attachment = new Attachment({
  filename: 'file.txt',
  type: 'plain/text',
  disposition: 'attachment',
  contentId: 'mytext',
  file: {
    path: '~/file.txt',
    encoding: 'utf8'
  }
});

sgMail.setApiKey(process.env.SENDGRID_API_KEY);
const msg = {
  to: 'name@example.com',
  from: 'name@example.com',
  subject: 'Hello attachment',
  html: '<p>Here’s an attachment for you!</p>',
  attachments: [
    attachment
  ],
};

sgMail.send(msg);
```

packages/helpers/classes/attachment.js
```diff
@@ -39,17 +40,29 @@ class Attachment {
     data = toCamelCase(data);
 
     //Extract properties from data
-    const {content, filename, type, disposition, contentId} = data;
+    const {content, filename, type, disposition, contentId, file} = data;
```
adamreisnz: I'm worried that `file` vs. `filename` will be a bit confusing. Maybe instead of passing a file object, we can have two separate new settings `filePath` and `fileEncoding`. 

(The `filename` setting should probably be deprecated and introduce a new one `fileName` to be consistent)

packages/helpers/classes/attachment.js
```diff
@@ -39,17 +40,29 @@ class Attachment {
...
+
+    let fileContent = !file ? content : this.readFile(file);
```
adamreisnz: File content is immutable so can be `const`. I'm not sure about the logic either though. If content is passed, should that take priority over a file? Maybe a warning if both are passed?

packages/helpers/classes/attachment.js
```diff
@@ -39,17 +40,29 @@ class Attachment {
...
     this.setFilename(filename);
     this.setType(type);
     this.setDisposition(disposition);
     this.setContentId(contentId);
   }
 
   /**
+   * Read a file and return its content as base64
+   */
+  readFile(file) {
```
adamreisnz: If we use a separate setting for the encoding you can pass it in here, with a default value of `utf8`, e.g.

```
readFile(path, encoding = 'utf8') {
   ...
}
```

packages/helpers/classes/attachment.js
```diff
@@ -39,17 +40,29 @@ class Attachment {
...
+    let encoding = !file.encoding ? 'utf8' : file.encoding;
+    let data = fs.readFileSync(file.path);
+    let buff = Buffer.from(data, encoding);
```
adamreisnz: All `let`'s are `const`ant and I'm wondering if we should return the buffer, instead of converting to string at this point? That way you can use the helper function to read files and get buffers and manipulate them if needed before passing to the attachment.

Also, we can update the Attachment class to detect if a Buffer is passed and automatically convert to string.

packages/helpers/classes/attachment.js
spelcaster: I think that the file should take priority over content, and warn the user if both is passed is a good idea.

packages/helpers/classes/attachment.js
spelcaster: Thanks for the tip.
Should I convert the method readFile to a helper function? Do you think that another helper function to convert a string to Buffer is necessary? 

According to [fs.readFileSync documentation](https://nodejs.org/api/fs.html#fs_fs_readfilesync_path_options) if the `encoding` is passed, then a string is returned, otherwise it'll return a buffer. 

Perhaps in this case the buffer conversion using an encoding is completely unnecessary, just if the user want to get a string from the buffer, then the encoding would be useful.

packages/helpers/classes/attachment.js
spelcaster: Is there any example of how to give the user a warning here?

adamreisnz: Actually we don't show warnings anywhere really, just throw errors. Maybe an error is suitable, because you are not expected to pass both values. Then, we also don't need to worry about what takes priority.

packages/helpers/classes/attachment.js
adamreisnz: Might not need to expose it, as it's basically just using `readFileSync`. Let's pass the encoding straight to `readFileSync`, so we won't need to convert to buffer.

packages/helpers/classes/attachment.js
```diff
@@ -40,38 +40,65 @@ class Attachment {
...
+      filename,
+      type,
+      disposition,
+      contentId,
+      filePath
+    } = data;
+
+    if ((typeof content !== 'undefined') && (typeof filePath !== 'undefined')) {
+      throw new Error(
+        'The props \'content\' and \'filePath\' cannot be used together.'
```
adamreisnz: Could you use template strings to make it a bit cleaner without the backslashes? E.g.

```
`The props 'content' and 'filePath' cannot be used together.`
```

packages/helpers/classes/attachment.js
```diff
@@ -40,38 +40,65 @@ class Attachment {
...
     this.setDisposition(disposition);
     this.setContentId(contentId);
+
+    if (filePath) {
+      const fileContent = this.readFile(filePath);
+      this.setContent(fileContent);
+      return;
+    }
+
+    this.setContent(content);
```
adamreisnz: Could use a one liner for all of this:

```
this.setContent(filePath ? this.readFile(filePath) : content);
```

packages/helpers/classes/attachment.js
```diff
@@ -40,38 +40,65 @@ class Attachment {
...
    */
   setContent(content) {
-    if (typeof content === 'undefined') {
+    const contentType = typeof content;
+
+    if (contentType === 'undefined') {
       return;
-    }
-    if (typeof content !== 'string') {
+    } else if (contentType === 'object'
```
adamreisnz: Could you please install ESLint for your code editor and run it on your code to ensure the code style is the same to the rest of the code base?

packages/helpers/classes/attachment.js
```diff
@@ -40,38 +40,65 @@ class Attachment {
...
+      && content.hasOwnProperty('type')
+      && (content.type === 'Buffer')
+    ) {
+      const buff = new Buffer(content.data);
+      this.content = buff.toString('base64');
+      return;
+    } else if (content instanceof Buffer) {
+      this.content = content.toString('base64');
+      return;
+    } else if (contentType !== 'string') {
```
adamreisnz: Could you write matching unit tests for all these new content type cases?

packages/helpers/classes/attachment.js
spelcaster: Sure, I was using the StandardJS

packages/helpers/classes/attachment.js
spelcaster: I will...

spelcaster:
@adamreisnz the changes were made, I'm sorry for the long response time, last week I was so busy.

packages/helpers/classes/attachment.js
adamreisnz: Have you used ESLint on your changes? The code style is still different.

adamreisnz:
@thinkingserious @mbernier I'm happy with this, assuming it works as intended and has been tested!

spelcaster:
cool! thanks for your review @adamreisnz :)

mbernier:
@spelcaster

We have not been able to merge your Pull Request, but because you are awesome - we wanted to make sure you could still get a SendGrid Hacktoberfest shirt.  

Please go **fill [out our swag form](https://dx.sendgrid.com/swag) before Nov 5th** and we will send the shirt! (We know that you might have tried this before and it didn’t work, sorry about that!)

### **You have till Nov 5th to fill out this form** in order to get the Hacktoberfest shirt!

Thank you for contributing during Hacktoberfest! We hope to see you in the repos soon! Just so you know, we always give away a SendGrid shirt for your first ever non-Hacktoberfest PR that gets merged.

mgadirocks:
@spelcaster I have tried your above code and tried to upload a 10mb pdf file. But It does not work. Again it does not work with a simple text file.

I have set the type to 'application/pdf' then too it does not work. Any help will be appriciated.
Refer the below code.


     ```//sample file
      const fs = require('fs');
     const sgMail = require('@sendgrid/mail');

     const {
         classes: {
             Attachment
         }
     } = require('@sendgrid/helpers');

     let attachment = new Attachment({
         filename: 'test.pdf',
         type: 'application/pdf',
         disposition: 'attachment',
         contentId: 'testingpdf',
         filePath: './attachments/test.pdf',
         //contentType:'application/pdf'
         //content: fs.readFileSync('./attachments/test.pdf'), // will work
         //content: 'w6cK', // will work
     });

     sgMail.setApiKey(process.env.SENDGRID_API_KEY);
     const msg = {
         to: 'example@gmail.com',
         from: 'abc@abc.com',
         subject: 'Hello attachment',
         html: '<p>Here’s an attachment for you!</p>',
         attachments: [
             attachment
         ],
     };

     sgMail.send(msg);

```

spelcaster:
@mgadirocks Have you tried with absolute path? See https://github.com/sendgrid/sendgrid-nodejs/pull/505/files#diff-de4f2f8e9851d3bd62c6424afddcf90fR58

clee:
LGTM

thinkingserious:
Hi @spelcaster,

I'm ready to merge this now, thanks for your patience!

Could you please take a moment to fix the merge conflicts?

Thank you!

With Best Regards,

Elmer

spelcaster:
@thinkingserious I've squashed the commits into only 1 and resolved the conflict. Could you verify if the rule to encode the buffer into a base64 string is correct? 

Thanks...

thinkingserious:
Hello @spelcaster,<br /><br/>Thanks again for the PR!<br />                                              <br />We appreciate your contribution and look forward to                                              continued collaboration. Thanks!                                             <br /><br />Team SendGrid DX

Issue #476, opened 2017-10-11
Cannot import package with TypeScript 2.5.x and 2.6.x


ShaharHD:
Trying to import the package in a Typescript file:
```import * as SendGrid from '@sendgrid/mail';```

gives the following error on tsc build:
```
src/libs/communications.ts(1,27): error TS2497: Module '"<long path>/node_modules/@sendgrid/mail/index"' resolves to a non-module entity and cannot be imported using this construct.
```

Any known resolution for this issue except using the following workaround?
`import SendGrid = require('@sendgrid/mail'); // tslint:disable-line`

**Comment by thinkingserious** [comment]
@SPARTAN563,

Any ideas on this one? Thanks!

**Comment by duncanleung** [comment]
Hi @thinkingserious,

It looks like this SO post addresses this issue:
https://stackoverflow.com/questions/39415661/what-does-resolves-to-a-non-module-entity-and-cannot-be-imported-using-this


From what I understand...

In `packages/mail/src/mail.d.ts` (https://github.com/sendgrid/sendgrid-nodejs/blob/master/packages/mail/src/mail.d.ts), it looks like the export is using CommonJS export, and not ES6 `export default`:

```
//packages/mail/src/mail.d.ts  line 33
export = new MailService()
```

Thus, you can't use ES6 imports as the OP desired:
```
import * as SendGrid from '@sendgrid/mail';
```

Instead, he needed to use CommonJS require. It's not a 'hack'. It's just following the same convention:
```
import SendGrid = require('@sendgrid/mail');
```

thinkingserious:
Thanks @duncanleung!

@ShaharHD,

Does that answer solve your issue?

ShaharHD:
Not really ... as my original post actually gave the workaround which was
`import SendGrid = require('@sendgrid/mail'); // tslint:disable-line`

I was asking to see if there's a cleaner way of doing it as in TS the standard is to have tslint to not allow `require` (hence the `// tslint:disable-line`)

Maybe this should be included in the README.md of the repo as the workaround for using sendgrid with TypeScript?

duncanleung:
@thinkingserious Referencing the same SO article:

> "If you control both modules, you can use export default instead"
https://stackoverflow.com/questions/39415661/what-does-resolves-to-a-non-module-entity-and-cannot-be-imported-using-this

The long-term fix and action item here, is to rewrite any CommonJS `export =` to use the ES6 `export default` in the sendgrid-nodejs lib

[...]

PR #606, opened 2017-11-06
fix: Address an issue with the TypeScript definitions for TS2.6+

SPARTAN563:
This PR addresses an issue raised by @ShaharHD in #476 which is caused by [TS2714](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#arbitrary-expressions-are-forbidden-in-export-assignments-in-ambient-contexts) - specifically that the definition files may not make use of expressions as part of their `export = ` statements.

Instead, this PR updates the definitions to use the suggested pattern of:
```
const x: X
export = x
```
rather than:
```
export = new X()
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "mocha-sinon": "^2.0.0",
     "sinon": "^2.3.2",
     "sinon-chai": "^2.10.0",
-    "typescript": "~2.4.2"
+    "typescript": "~2.6.1"
   },
   "scripts": {
     "lint": "eslint . --fix",
--- a/packages/client/src/client.d.ts
+++ b/packages/client/src/client.d.ts
@@ -34,4 +34,5 @@ declare class Client {
   request(data: ClientRequest, cb?: (err: ResponseError, response: [ClientResponse, any]) => void): Promise<[ClientResponse, any]>;
 }
 
-export = new Client()
+declare const client: Client;
+export = client;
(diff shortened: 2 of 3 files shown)


ShaharHD:
is there an ETA on this PR?

thinkingserious:
Hi @ShaharHD,

We don't have a firm time on the ETA, but I can give you some insight on our process. 

Firstly, we are currently behind on merging PRs due to an amazing influx of PRs for Hacktoberfest. We had over 1000!

Generally, PRs rise in our backlog based on several factors, but the one factor the community can influence is what you did. Either comment or add a +1 to the thread. So thank you for that!

We appreciate your support and patience.

With Best Regards,

Elmer

ShaharHD:
Hi @thinkingserious - thank you for taking the time to response.

We've locked our projects using SendGrid to TS `"~2.5"` in the time being

phra:
i suggest to close this in favor of #619 because it fixes problems with `strict: true` environment. i didn't bumped the version of typescript of the package. i can do it if required.

nolazybits:
Trying to use your library with Typescript and it is painful for the issue mentioned in this PR but also 
trying to have a var of type MailService, but this is not possible as not the class is not exported.
Maybe part of this PR could you please review your definition files and be sure we can have full type checking please.

Thanks

SPARTAN563:
@nolazybits #619 was merged instead of this PR and it includes the ability to consume the `MailService` type if you need it. Please let us know on #619 if this isn't working as intended for you.

PR #868, opened 2019-01-03
Fix #856 - dynamicTemplateData should support object and array, not only strings

### LINKED ISSUE #856: dynamicTemplateData should support object and array, not only strings
#### Issue Summary

MailData type support only string values, while Handlebars support JSON.
In the Transactional Templates it's possible to have "Hello, {{obj.field}}!".
I tried myself using Test Data in Transactional Templates Preview. Also it is stated in the official documentation at https://sendgrid.com/docs/ui/sending-email/how-to-send-an-email-with-dynamic-transactional-templates/ and https://sendgrid.com/docs/ui/sending-email/using-handlebars/

In MailData instead, the field personalizations.dynamicTemplateData has type { [key: string]: string; };
As values we are limited to string, making it practically impossible to code conditionals or multi language templates. It also force to update the code whenever we want to add a field in the template, while we could simply pass the entire object.

* sendgrid-nodejs Version: 6.3.1

thinkingserious:
Fixes #856

### DIFF
--- a/packages/helpers/classes/personalization.d.ts
+++ b/packages/helpers/classes/personalization.d.ts
@@ -7,7 +7,7 @@ export interface PersonalizationData {
   subject?: string;
   headers?: { [key: string]: string };
   substitutions?: { [key: string]: string };
-  dynamicTemplateData?: { [key: string]: string; };
+  dynamicTemplateData?: { [key: string]: object; };
   customArgs?: { [key: string]: string };
   sendAt?: number;
 }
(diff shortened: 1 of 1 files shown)


thinkingserious:
@SPARTAN563 @adamreisnz,

A code review would be much appreciated. Thanks!

With Best Regards,

Elmer

SPARTAN563:
I suspect you're going to want to use something like this instead: `object | string` but I need to look closer at the exact values which are accepted before I can fully review this and say one way or another. Will try and get to that in an hour or so.

SPARTAN563:
Looks to me like you accept just about any valid JSON object there (reading from [here](https://sendgrid.com/docs/ui/sending-email/how-to-send-an-email-with-dynamic-transactional-templates/)) so I'd lean towards using the following:

```typescript
dynamic_template_data?: { [key: string]: any }
```

That will allow the inclusion of any valid JS object or value and while it would (in theory) accept functions and other types, the JSON serializer should ignore those unless they have a `.toJSON()` method.

childish-sambino:
Duplicates #939

PR #892, opened 2019-02-20
Dynamic template data and substitutions not appended to JSON

adamreisnz:
This fixes dynamic template data and substitutions not getting appended to JSON requests.

### Checklist
- [x] I have read the [Contribution Guide] and my PR follows them.
- [x] I updated my branch with the development branch.
- [ ] I have added necessary documentation about the functionality in the appropriate .md file
- [x] I have added in line documentation to the code I modified

### Short description of what this PR does:
I found that dynamic template data was not sent at all with any request, even if using a dynamic template ID. This PR fixes that and ensures the data is sent to the server.

### DIFF
--- a/packages/helpers/classes/mail.js
+++ b/packages/helpers/classes/mail.js
@@ -575,9 +575,6 @@ class Mail {
     if (Object.keys(trackingSettings).length > 0) {
       json.trackingSettings = trackingSettings;
     }
-    if (substitutions && Object.keys(substitutions).length > 0) {
-      json.substitutions = substitutions;
-    }
     if (dynamicTemplateData && Object.keys(dynamicTemplateData).length > 0) {
       json.dynamicTemplateData = dynamicTemplateData;
     }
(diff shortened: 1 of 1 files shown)


adamreisnz:
@thinkingserious any chance you could expedite this fix?

childish-sambino:
This doesn't look valid to me. `dynamicTemplateData` defined at the top-level gets applied to the inner personalizations. Including them in the outer mail JSON is therefore not necessary.

adamreisnz:
Why was this closed without discussion? I am the one that developed this package, so maybe it's sensible to have a chat about it first?

From what I recall, this was necessary for things like the subject etc. to be correctly parsed using dynamic template data.

adamreisnz:
cc @thinkingserious @childish-sambino

childish-sambino:
Just tried sending a dynamic template with a templated subject and didn't have an issue. I don't see this change as being necessary. The top-level template data gets merged into the personalizations.

adamreisnz:
Ok, I appreciate your looking into this. As you can see, this PR was created about a year ago, so I can't recall anymore exactly the use case or what it was needed for. I will have to dig into it deeper and make some examples to see why I thought it was necessary.

Also, this PR contained some other changes, so I would appreciate re-opening it and I can review it when I have a moment and if needed drop the top level data but keep the other changes.

PR #1041, opened 2020-02-14
fix: require in TS at least 1 type of content for sending mail

### LINKED ISSUE #978: TypeScript types allow mails with missing bodies
#### Issue Summary

Each SendGrid mail requires either a `text` key or an `html` key. If neither is present, the server will reject the mail.

The SendGrid TypeScript types' `MailData` type is the type passed to `sendgrid.send`. It contain [this](https://github.com/sendgrid/sendgrid-nodejs/blob/c134ecd09080fb9d5906a8b62b80db93e6816d56/packages/helpers/classes/mail.d.ts#L117):

```
export interface MailData {
  /* other keys elided */
  text?: string,
  html?: string,
  /* other keys elided */
}
```

This is wrong, because it means that both `text` and `html` are allowed to be absent. If you send such a request to the server, it will always be rejected with HTTP 400 Bad Request.

#### Steps to Reproduce

1. Install the `sendgrid/mail` NPM package.
2. Write TypeScript code that calls `sendgrid.send({from, to, subject})` (with no `text` or `html` provided).
3. That code type checks.
4. Run that code with a valid SendGrid API key.
5. The server will always reject it because it has no body.

This is a bug in the types: they allow me to do something that's (1) invalid, and (2) statically preventable in TypeScript's type system. Our user feedback form was down in production for 24 hours because of this bug.

#### Technical details:

This can be fixed by doing something like:

```
export interface MailData {
  /* other keys elided */
} & (
  {text: string} | {html: string} | {text: string, html: string}
)
```

With that change, the user can no longer call `sendgrid.send` without providing at least one type of body. (There are probably more types of bodies that I don't know about, and which would need to be considered in the union as well. I assume that's what the `content` key is, for example. But I only know a relatively small part of SendGrid's API.)

childish-sambino:
Fixes #978

### DIFF
--- a/packages/mail/src/mail.d.ts
+++ b/packages/mail/src/mail.d.ts
@@ -1,7 +1,9 @@
-import Client = require("@sendgrid/client");
 import {ClientResponse} from "@sendgrid/client/src/response";
 import {ResponseError} from "@sendgrid/helpers/classes";
-import {MailData} from "@sendgrid/helpers/classes/mail";
+import {MailContent, MailData} from "@sendgrid/helpers/classes/mail";
+
+type MailDataRequired = MailData & (
+  { text: string } | { html: string } | { content: MailContent[] & { 0: MailContent } });
 
 declare class MailService {
   /**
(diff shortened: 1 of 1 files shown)


thinkingserious:
Hello @childish-sambino,<br /><br/>Thanks again for the PR!<br />                                              <br />We appreciate your contribution and look forward to                                              continued collaboration. Thanks!                                             <br /><br />Team SendGrid DX

JClackett:
This has introduced a TS error when sending mails with just templateIds: 

```ts

const data = {
	from: "someemail",
    to: "somemail",
    templateId: "some sendgrid templateId",
    dynamic_template_data: {randonVariables: "something"},
}

sendgrid.send(data)

```

```
Property 'content' is missing in type '{ from: string; to: string | string[]; templateId: string; dynamic_template_data: any; }' but required in type '{ content: MailContent[] & { 0: MailContent; }; }'.
```
We dont require there to be a body or content as its using a template?

Any ideas?

childish-sambino:
🤦‍♂ Need more `|`s.

krazibit:
as @JClackett mentioned, template use case was not included.
anyone working to add the type union in the MailDataRequired already?

childish-sambino:
@krazibit It was included as part of #1053

krazibit:
great, thanks!

PR #1053, opened 2020-03-05
chore: export type MailDataRequired

antonsamper:
Exposes the new mail properties type `MailDataRequired` that was added in https://github.com/sendgrid/sendgrid-nodejs/pull/1041

### Checklist
- [x] I acknowledge that all my contributions will be made under the project's license
- [x] I have made a material change to the repo (functionality, testing, spelling, grammar)
- [x] I have read the [Contribution Guidelines](CONTRIBUTING.md) and my PR follows them
- [x] I have titled the PR appropriately
- [x] I have updated my branch with the master branch
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation about the functionality in the appropriate .md file
- [ ] I have added inline documentation to the code I modified

PR #1054, opened 2020-03-05
fix: allow template ID and no content in TS when sending mail

childish-sambino:
Fixes issue here: https://github.com/sendgrid/sendgrid-nodejs/pull/1041#issuecomment-595332191


childish-sambino:
Duplicates https://github.com/sendgrid/sendgrid-nodejs/pull/1053

PR #1058, opened 2020-03-09
feat!: migrate from deprecated request module to axios

### LINKED ISSUE #1042: warning > request@2.88.2 deprecated
warning @sendgrid/mail > @sendgrid/client > request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142

eshanholtz:
Fixes #1042


eshanholtz:
Upgrading `eslint` (security remediation) requires node >= 8

packages/client/src/classes/client.js
```diff
@@ -112,21 +122,25 @@ class Client {
...
-        //Request error
-        if (error) {
+      axios(request)
+        .then(response => {
+          // Successful response
+          var parsedResponse = {
+            statusCode: response.status,
+            body: response.data,
+          };
+          return resolve([parsedResponse, response.data]);
```
childish-sambino: This is a potentially breaking change so I'd like some feedback if we think there should be a major version roll.

Previously, this returned `[RequestResponse, any]` and now it's returning `[{body: any, statusCode: number}, any]` which does not match the TS which is `[AxiosResponse, any]`. Ideally, this should not be exposing the underlying type because it's an implementation detail that may change in the future. Similar to the request type.

I think the types should be defined in this repo and not expose the Axios/Request lib types.

package.json
```diff
@@ -14,7 +14,7 @@
     "chai": "^4.0.1",
     "chai-as-promised": "^7.1.1",
     "dirty-chai": "^2.0.1",
-    "eslint": "^4.19.1",
+    "eslint": "^6.8.0",
```
childish-sambino: I don't like having to drop support for versions of Node in order to support a dev dependency. Prefer to get some usage numbers first to determine if it's okay to drop these first.

packages/client/src/classes/client.js
eshanholtz: I agree. I'll implement an interface similar to [what we have in twilio-node](https://github.com/twilio/twilio-node/tree/master/lib/http). I also agree that this should warrant a major version roll

package.json
childish-sambino: Can we use a 5.X version for now and/or move this to another PR?

packages/helpers/classes/index.d.ts
```diff
@@ -2,12 +2,16 @@ import Attachment from "@sendgrid/helpers/classes/attachment";
 import EmailAddress from "@sendgrid/helpers/classes/email-address";
 import Mail from "@sendgrid/helpers/classes/mail"
 import Personalization from "@sendgrid/helpers/classes/personalization";
+import Request from "@sendgrid/helpers/classes/request";
 import ResponseError from "@sendgrid/helpers/classes/response-error";
+import Response from "@sendgrid/helpers/classes/response";
```
childish-sambino: nit: `Response` before `ResponseError` here and elsewhere.

packages/client/src/client.d.ts
```diff
@@ -26,12 +29,49 @@ declare class Client {
...
 
   /**
    * Do a request
    */
-  request(data: ClientRequest, cb?: (err: ResponseError, response: [ClientResponse, any]) => void): Promise<[ClientResponse, any]>;
+  request<TData>(opts: Client.RequestOptions<TData>, cb?: (err: ResponseError, response: [ClientResponse, any]) => void): Promise<[ClientResponse, any]>;
+}
+
+declare namespace Client {
+  export interface RequestOptions<TData = any, TParams = object> {
```
childish-sambino: This looks like mostly a duplicate of `Request.RequestOptions`.

packages/helpers/classes/request.js
```diff
@@ -0,0 +1,77 @@
+'use strict';
+
+var _ = require('lodash');
```
childish-sambino: Generally we're trying to move away from using lodash. Should eval if this can be dropped.

test.js
```diff
@@ -0,0 +1,15 @@
+
+// using Twilio SendGrid's v3 Node.js Library
+// https://github.com/sendgrid/sendgrid-nodejs
+const sgMail = require('./packages/mail');
+sgMail.setApiKey(process.env.SENDGRID_API_KEY);
+const msg = {
+  to: 'ea.shanholtz@gmail.com',
+  from: 'ea.shanholtz@gmail.com',
```
childish-sambino: Don't think this belongs.

package.json
eshanholtz: Done: https://github.com/sendgrid/sendgrid-nodejs/pull/1063/files

packages/client/src/client.d.ts
eshanholtz: So I've played around with this a bit, and haven't been able to import Request.RequestOptions. We have a similar duplication in `twilio-node`: [request.d.ts](https://github.com/twilio/twilio-node/blob/master/lib/http/request.d.ts) vs. [RequestClient.d.ts](https://github.com/twilio/twilio-node/blob/master/lib/base/RequestClient.d.ts)

eshanholtz: Figured out a workaround

package.json
```diff
@@ -14,7 +14,7 @@
     "chai": "^4.0.1",
     "chai-as-promised": "^7.1.1",
     "dirty-chai": "^2.0.1",
-    "eslint": "^4.19.1",
+    "eslint": "^5.16.0",
```
childish-sambino: Should be part of a separate PR.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: node_js
 node_js:
   - 6
-  - 7
```
childish-sambino: Should be part of a separate PR.

packages/client/src/classes/client.js
```diff
@@ -72,7 +74,7 @@ class Client {
   createHeaders(data) {
 
     //Merge data with default headers
-    const headers = mergeData(this.defaultHeaders, data);
+    let headers = mergeData(this.defaultHeaders, data);
```
childish-sambino: Why this change?

packages/client/src/classes/client.js
```diff
@@ -86,47 +88,53 @@ class Client {
...
-
-    //Keep URL parameter consistent
-    if (data.uri) {
-      data.url = data.uri;
-      delete data.uri;
-    }
+  createRequest(opts) {
+
+    let options = {
+      url: opts.url || opts.uri,
```
childish-sambino: For compatibility with above deleted block, should be swapped:
```
opts.uri || opts.url
```

packages/client/src/classes/client.js
```diff
@@ -86,47 +88,53 @@ class Client {
...
+      baseURL: opts.baseURL || this.defaultRequest.baseURL,
+      method: opts.method || this.defaultRequest.method,
```
childish-sambino: Isn't this what `mergeData` does?

packages/client/src/classes/client.js
```diff
@@ -86,47 +88,53 @@ class Client {
...
     const promise = new Promise((resolve, reject) => {
-      http(request, (error, response, body) => {
-
-        //Request error
-        if (error) {
+      axios(opts)
+        .then(response => {
+          // Successful response
+          _this.lastResponse = new Response(response.status, response.data, response.headers);
+          return resolve([_this.lastResponse, response.data]);
```
childish-sambino: This could result in threading issues if `lastResponse` get reassigned before returning.

packages/helpers/classes/attachment.js
```diff
@@ -79,8 +79,7 @@ class Attachment {
     if (typeof content === 'string') {
       this.content = content;
       return;
-    }
-    else if (content instanceof Buffer && content.toString !== undefined) {
+    } else if (content instanceof Buffer && content.toString !== undefined) {
```
childish-sambino: Should be part of a separate PR.

packages/helpers/classes/attachment.js
```diff
@@ -52,7 +52,7 @@ class Attachment {
 
     if ((typeof content !== 'undefined') && (typeof filePath !== 'undefined')) {
       throw new Error(
-        `The props 'content' and 'filePath' cannot be used together.`
+        'The props \'content\' and \'filePath\' cannot be used together.'
```
childish-sambino: Should be part of a separate PR.

packages/helpers/classes/index.d.ts
```diff
@@ -2,12 +2,16 @@ import Attachment from "@sendgrid/helpers/classes/attachment";
...
   Attachment,
   EmailAddress,
   Mail,
   Personalization,
-  ResponseError,
-}
\ No newline at end of file
+  Request,
+  Response,
+  ResponseError
```
childish-sambino: nit: Keep trailing `,`

packages/helpers/classes/request.d.ts
```diff
@@ -0,0 +1,15 @@
+type HttpMethod = 'get'| 'GET'|'post'|'POST'|'put'|'PUT'|'patch'|'PATCH'|'delete'|'DELETE';
```
childish-sambino: nit: Spacing

packages/helpers/classes/request.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+class Request {
+  constructor(opts) {
+    opts = opts || {};
+    this.ANY = '*';
```
childish-sambino: What's the point/value of this?

package.json
eshanholtz: It is: https://github.com/sendgrid/sendgrid-nodejs/pull/1063

packages/client/src/classes/client.js
eshanholtz: Ah. Must have forgotten to put it back. Just playing around with the changes.

packages/helpers/classes/request.js
eshanholtz: It was used for Request comparison, but since I took that out with lodash, I should take this out, too.

.travis.yml
childish-sambino: What I mean is the changes should be excluded from this PR.

packages/client/src/classes/client.js
eshanholtz: I think I fixed this problem, can you double check?

packages/client/src/classes/client.js
```diff
@@ -3,22 +3,22 @@
...
-const http = require('request');
+const axios = require('axios');
 const pkg = require('../../package.json');
 const {
   helpers: {
     mergeData,
   },
   classes: {
     ResponseError,
+    Response,
```
childish-sambino: nit: ordering

packages/client/src/classes/client.js
```diff
@@ -30,13 +30,14 @@ class Client {
...
+      'Content-Type': 'application/json',
       'User-agent': 'sendgrid/' + pkg.version + ';nodejs',
     };
 
     //Empty default request
     this.defaultRequest = {
-      json: true,
-      baseUrl: 'https://api.sendgrid.com/',
+      data: {},
+      baseURL: 'https://api.sendgrid.com/',
```
childish-sambino: Prefer to make this `baseUrl` in the default request and then adapt it later for axios.

packages/client/src/classes/client.js
```diff
@@ -62,7 +63,11 @@ class Client {
    * Set default request
    */
   setDefaultRequest(key, value) {
-    this.defaultRequest[key] = value;
+    if (key === 'baseUrl') {
```
childish-sambino: Then this logic isn't needed.

packages/client/src/classes/client.js
```diff
@@ -86,47 +91,52 @@ class Client {
...
-    }
+  createRequest(opts) {
+
+    let options = {
+      url: opts.uri || opts.url,
+      baseURL: opts.baseUrl,
+      method: opts.method,
+      data: opts.body || {},
+      params: opts.qs || {},
+      headers: opts.headers || {},
```
childish-sambino: Is this `||`ing here needed?

packages/client/src/request.d.ts
```diff
@@ -1,3 +1,4 @@
-import {OptionsWithUrl} from "request";
+import { Request, RequestOptions } from "@sendgrid/helpers/classes/request";
```
childish-sambino: nit: Spacing around curly brackets is not consistent between files.

packages/helpers/classes/response-error.js
```diff
@@ -19,12 +19,12 @@ class ResponseError extends Error {
...
-    const {headers, statusCode, statusMessage, body} = response;
+    const {headers, status, statusText, data} = response;
 
     //Set data
-    this.code = statusCode;
-    this.message = statusMessage;
-    this.response = {headers, body};
+    this.code = status;
+    this.message = statusText;
+    this.response = {headers, data};
```
childish-sambino: Should be `{headers, body: data}` for compatibility.

packages/helpers/classes/request.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+class Request {
```
childish-sambino: Is this class still needed?

packages/mail/src/classes/mail-service.js
```diff
@@ -192,7 +192,7 @@ class MailService {
       const request = {
         method: 'POST',
         url: '/v3/mail/send',
-        body,
+        data: body,
```
childish-sambino: Don't think this should change.

test/typescript/client.ts
```diff
@@ -26,5 +26,5 @@ const req = Client.createRequest({
 Client.request({
   url: "/test"
 }).then(res => {
-  res[0].headers
-})
\ No newline at end of file
+  res[0].statusCode;
```
childish-sambino: From `headers` to `statusCode`?

packages/helpers/classes/request.js
eshanholtz: I guess we only need the typescript one

test/typescript/client.ts
eshanholtz: Shouldn't matter, I was making sure the typescript response fields were populating properly. and that I had the right variable names.

packages/client/src/classes/client.js
eshanholtz: Where is an appropriate "later"? Also, if this dict is meant to be merged with the renamed opts in the `request` function, then the `defaultRequest` should have keys that align with what's adapted for axios

packages/client/src/classes/client.js
eshanholtz: Nope

packages/client/src/classes/client.js
eshanholtz: @childish-sambino can I get some feedback?

childish-sambino: I think it should be adapted after L105 where the data is merged. We define the structure of default request, merge it with the options provided in the request, and then adapt it to be whatever the underlying client is expecting.

packages/client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@sendgrid/client",
   "description": "Twilio SendGrid NodeJS API client",
-  "version": "6.5.5",
+  "version": "6.5.3",
```
childish-sambino: Needs to be sync'd.

packages/helpers/classes/response-error.js
```diff
@@ -40,10 +40,10 @@ class ResponseError extends Error {
    * Convert to string
    */
   toString() {
-    const {body} = this.response;
+    const {data} = this.response;
```
childish-sambino: This should be reverted back to `body`.

packages/helpers/classes/request.d.ts
```diff
@@ -0,0 +1,15 @@
...
+    method?: HttpMethod | '*';
+    url?: string;
+    auth?: string;
+    params?: TParams;
+    data?: TData | '*';
+    headers?: object | '*';
+    ca?: string;
+}
+
+export class Request<TData = any> {
```
childish-sambino: This class doesn't exist anymore in the JS.

packages/helpers/classes/request.d.ts
```diff
@@ -0,0 +1,15 @@
+type HttpMethod = 'get'|'GET'|'post'|'POST'|'put'|'PUT'|'patch'|'PATCH'|'delete'|'DELETE';
+
+export interface RequestOptions<TData = any, TParams = object> {
+    method?: HttpMethod | '*';
+    url?: string;
+    auth?: string;
+    params?: TParams;
+    data?: TData | '*';
+    headers?: object | '*';
+    ca?: string;
```
childish-sambino: Should only contain `url`, `baseUrl`, `method`, `body`, `qs`, and `headers`.

packages/helpers/classes/response-error.js
eshanholtz: This uses the raw response object from `axios`

childish-sambino: No, it's using the response object defined in L27.

packages/client/src/request.d.ts
```diff
@@ -1,3 +1,3 @@
-import {OptionsWithUrl} from "request";
+import RequestOptions from "@sendgrid/helpers/classes/request";
 
-export type ClientRequest = OptionsWithUrl;
\ No newline at end of file
+export default RequestOptions;
```
childish-sambino: Can we just keep this a `ClientRequest` for compatibility (and less changes)?

eshanholtz: Done, also put back some other unnecessary changes

jandppw:
Can somebody specify _why_ this is breaking? (So we can check whether it has impact on our usage code). I can't see that from browsing the code changes.

childish-sambino:
@jandppw https://github.com/sendgrid/sendgrid-nodejs/blob/master/docs/migration-guides/migrating-from-version-6-to-7.md

PR #1063, opened 2020-03-12
chore!: upgrade dev dependencies and fix eslint warnings/errors

eshanholtz:


.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: node_js
 node_js:
   - 6
-  - 7
```
eshanholtz: upgrading to eslint 5.16.0 required to fix security vulnerability does not support node 7

packages/helpers/classes/attachment.js
```diff
@@ -52,7 +52,7 @@ class Attachment {
 
     if ((typeof content !== 'undefined') && (typeof filePath !== 'undefined')) {
       throw new Error(
-        `The props 'content' and 'filePath' cannot be used together.`
+        'The props \'content\' and \'filePath\' cannot be used together.'
```
childish-sambino: nit: Just wrap the string in double-quotes instead of backticks.

.travis.yml
childish-sambino: Let's mark this as breaking and just move to `6.X` for eslint. We'll hold off on merging until after knowing usage data.

PR #1069, opened 2020-03-18
chore: refactor duplicate code in client tests

childish-sambino:


packages/client/src/client.spec.js
```diff
@@ -1,83 +1,68 @@
...
+    return testRequest(request, 200);
   });
 });
 
 describe('test_access_settings_whitelist_post', () => {
   const request = {};
-  const data = {
-    'ips': [
+  request.body = {
+    ips: [
```
eshanholtz: @childish-sambino did you remove quotes manually or did you use a tool/linter?

childish-sambino: Told IntelliJ to fix all occurrences in the file.

PR #1077, opened 2020-03-30
feat: console warning added when using an invalid API key

### LINKED ISSUE #1068: Error handling in setApiKey
### Issue Summary
I'm retrieving the api key from an environment variable. Somehow in one of our environments the api key was `undefined`. It would be helpful if an error is logged or an error is thrown when setupApiKey is called with `undefined`, or with other wrong inputs, so that we are immediately informed when this is the case.

### Steps to Reproduce
1. Setup the api key as undefined `sgMail.setupApiKey(undefined)`
2. No error, so it seems like the ApiKey is set correctly
3. Only when trying to send a mail an error is thrown

### Code Snippet
```node
const msg = {
  to: 'test@example.com',
  from: 'test@example.com',
  subject: 'Sending with SendGrid is Fun',
  text: 'and easy to do anywhere, even with Node.js',
  html: '<strong>and easy to do anywhere, even with Node.js</strong>',
};
sgMail.setApiKey(undefined);
sgMail.send(msg);
```

### Exception/Log

None after `setupApiKey(undefined)`

Error after sending the mail;
```
Unauthorized (401)
  Permission denied, wrong credentials
```

### Technical details:
* sendgrid-nodejs version: 6.5.4
* node version: 8.16.2

childish-sambino:
Fixes #1068

### DIFF
--- a/packages/client/src/classes/client.js
+++ b/packages/client/src/classes/client.js
@@ -35,7 +35,7 @@ class Client {
     this.apiKey = apiKey;
 
     if (!this.isValidApiKey(apiKey)) {
-      console.warn(`API key does not start with ${API_KEY_PREFIX}`);
+      console.warn(`API key does not start with "${API_KEY_PREFIX}".`);
     }
   }
 
@@ -109,7 +109,7 @@ class Client {
         });
     });
 
-    // Throw an error in case a callback function was not passed/
+    // Throw an error in case a callback function was not passed.
     if (cb && typeof cb !== 'function') {
       throw new Error('Callback passed is not a function.');
     }
--- a/packages/client/src/classes/client.js
+++ b/packages/client/src/classes/client.js
@@ -1,8 +1,4 @@
 'use strict';
-
-/**
- * Dependencies
- */
 const axios = require('axios');
 const pkg = require('../../package.json');
 const {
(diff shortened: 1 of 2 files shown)


packages/client/src/classes/client.js
```diff
@@ -15,26 +11,18 @@ const {
...
-   */
   constructor() {
-
-    //API key
     this.apiKey = '';
 
-    //Default headers
     this.defaultHeaders = {
-      'Accept': 'application/json',
+      Accept: 'application/json',
```
thinkingserious: Why do we convert to a non-string key here? Why only for the `Accept` key?

packages/client/src/classes/client.js
```diff
@@ -43,51 +31,45 @@ class Client {
...
   }
 
-  /**
-   * Set API key
-   */
   setApiKey(apiKey) {
     this.apiKey = apiKey;
+
+    if (!this.isValidApiKey(apiKey)) {
+      console.warn(`API key does not start with ${API_KEY_PREFIX}`);
```
thinkingserious: Here is a spot to put the message: https://github.com/sendgrid/sendgrid-nodejs/blob/master/packages/helpers/constants/index.js -- nit: period at the end.

packages/client/src/classes/client.js
```diff
@@ -106,52 +88,40 @@ class Client {
...
               return reject(new ResponseError(error.response));
             }
           }
-          // Request error
           return reject(error);
         });
     });
 
-    // Throw and error incase function not passed
+    // Throw an error in case a callback function was not passed/
```
thinkingserious: nit: period for consistency

packages/client/src/classes/client.js
childish-sambino: Quotes aren't needed here and is an eslint warning.

packages/client/src/classes/client.js
childish-sambino: That's in the helpers package. I don't like defining the message in 1 package and using it in another.

PR #1136, opened 2020-06-03
feat: verify signature from event webhook

eshanholtz:
When enabling the "Signed Event Webhook Requests" feature in Mail Settings, Twilio SendGrid will generate a private and public key pair using the Elliptic Curve Digital Signature Algorithm (ECDSA). Once that is successfully enabled, all new event posts will have two new headers: X-Twilio-Email-Event-Webhook-Signature and X-Twilio-Email-Event-Webhook-Timestamp, which can be used to validate your events.

This SDK update will make it easier to verify signatures from signed event webhook requests by using the verifySignature method. Pass in the public key, event payload, signature, and timestamp to validate. Note: You will need to convert your public key string to an elliptic public key object in order to use the verifySignature method.


packages/client/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "main": "index.js",
   "engines": {
-    "node": ">=6.0.0"
+    "node": "6.* || 8.* || >=10.*"
```
childish-sambino, line 27: READMEs need updating. E.g., https://github.com/sendgrid/sendgrid-nodejs/blob/master/packages/mail/README.md#prerequisites

packages/mail/README.md
```diff
@@ -13,7 +13,7 @@ To be notified when this package is updated, please subscribe to email [notifica
 
 ## Prerequisites
 
-- Node.js version 6, 7 or 8
+- Node.js version 6, 8 or 8>=10
```
childish-sambino, line 16: New emoji?

eshanholtz: its a bird balancing on a skateboard on a ball

packages/eventwebhook/src/eventwebhook.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+const ecdsa = require('@starkbank/ecdsa');
+const Ecdsa = ecdsa.Ecdsa;
+const Signature = ecdsa.Signature;
+const PublicKey = ecdsa.PublicKey;
+
+class EventWebhook {
+  /**
+   *
```
childish-sambino, line 10: Missing summaries.

eshanholtz: Adding now. Should I submit a separate PR?

eshanholtz: done.

Issue #1128, opened 2020-05-20
Enable sandbox mode for all requests


malimccalla:
I have the following file that handles my sendgrid set up.

```ts
import sgMail from '@sendgrid/mail';

const sendGridApiKey = process.env.SENDGRID_API_KEY;

if (!sendGridApiKey) {
  throw new Error('The sendgrid api key has not been set in the environment variables');
}

sgMail.setApiKey(sendGridApiKey);

export { sgMail };
```

After setting my api key I need to be able to do something like this to enable sandbox mode on all requests

```ts
sgMail.enableSandbox(process.env.NODE_ENV !== 'production')
```

The absolute best solution would be not having to add a line of code like that at all but be able to generate an api key specifically for sandbox mode (similar to how Stripe has test keys). I really don't want to use my live production key in staging and development environments. Thanks!

malimccalla:
I can see this was marked as non-library issue. Where is the correct place to open this issue?

thinkingserious:
Hello @malimccalla,

We will route this request to the proper location internally on your behalf and update you here on any progress. Thank you!

With best regards,

Elmer

thinkingserious:
Internal tracking#: CL-2681

jasonribble:
Is anyone working on this?

mattiasjback:
I want this feature as well. Any plans on implementing it soon?

neo-geo15:
In case anyone else runs across this, a solution could be to implement a server which mocks the appropriate sendgrid endpoints. The example setup for the mock endpoint:

```
sgClient.setApiKey(SOME_KEY);
sgClient.setDefaultRequest('baseUrl', 'MOCK_SERVER_ENDPOINT');
sgMail.setClient(sgClient);
```

The point being that the library will run all its validation logic (as it would in sandbox mode), and one can simply return dummy data/results from the mock server all while ignoring the api key (since the mock server doesnt have to validate the key).

But I am just doing this for unit tests (and rather simple use cases) and I therefore am not sure how it would scale to bigger dev/stage environments with more complex setups in which more complex and accurate server responses are required.

PR #1169, opened 2020-07-31
Feature: Allow default data to be sent on all requests

seromenho:
Resolves: #1128  
👋
With this PR you can set default data you would like to send on each request. Eg: sandbox mode, from, etc..
I think this would be helpful, let me know what do you think.

### Usage
```js
sgMail.setApiKey(config.get('SENDGRID.API_KEY'))
sgMail.setDefaultData({
  from: 'always@from.me',
  mailSettings:  {
    sandboxMode: {
      enable: config.get('NODE_ENV') !== 'production'
    }
  }
});
```

### Notes
1. Naming might not be the best, please let me know.
2. Used deepmerge but I've seen there's merge-data-deep on helpers although not used and there's also deepmerge dependency on helpers. 

### Checklist
- [ ] I acknowledge that all my contributions will be made under the project's license
- [ ] I have read the [Contribution Guidelines](CONTRIBUTING.md) and my PR follows them
- [ ] I have titled the PR appropriately
- [ ] I have updated my branch with the master branch
- [ ] I have added necessary documentation about the functionality in the appropriate .md file


packages/mail/package.json
```diff
@@ -28,7 +28,8 @@
   },
   "dependencies": {
     "@sendgrid/client": "^7.2.2",
-    "@sendgrid/helpers": "^7.2.0"
+    "@sendgrid/helpers": "^7.2.0",
+    "deepmerge": "^4.2.2"
```
eshanholtz, line 32: I don't like the idea of adding another dependency to maintain this change. Can this be done without the additional package?

packages/mail/src/classes/mail-service.js
```diff
@@ -110,6 +112,17 @@ class MailService {
     });
   }
 
+  /**
+   * Set default data to for all requests. Can be overwritten by incoming data on send
+   */
+  setDefaultData(data) {
+    if (Object.prototype.toString.call(data) !== "[object Object]") {
```
eshanholtz, line 119: What is this doing?

packages/helpers/classes/mail.d.ts
```diff
@@ -119,7 +119,7 @@ export interface MailData {
   cc?: EmailData|EmailData[],
   bcc?: EmailData|EmailData[],
 
-  from: EmailData,
+  from?: EmailData,
```
eshanholtz, line 122: I'm not sure I like this change to the interface. This appears as if the `from` field is not required in a `MailData` payload, which is misleading.

packages/mail/src/classes/mail-service.js
seromenho: Checking data is an object.

packages/mail/package.json
seromenho: I thought on this. Left on the notes. I see this is also used on helpers, and doesn't make sense to export it from there to use it here. I could have used it even without adding it to the dependencies but I think have it explicit is better.
We can either use merge-data-deep from the helpers or not sure what's your approach to use `deepmerge`
Thoughts?

packages/helpers/classes/mail.d.ts
seromenho: I see your point and I agree. I saw there is validations on creating the Mail message. This would be caught on execution, got excited and tried to solve all in one 😅

seromenho:
Btw, I was also looking for something like this because I was trying to create a wrapper to the `.send` with some defaults. My issue was.... typescript 😅 It was being a bit tricky create this wrapper with typescript because both signatures have 3 params. It is picking the last signature send method where ideally, for my usage,  I would like to have them combined or use the first  one. (Works fine though using the second signature)

seromenho:
Just created https://github.com/sendgrid/sendgrid-nodejs/pull/1172 and this one might not be needed.

childish-sambino:
@seromenho I'm approving and merging #1172. Do you still want/need these changes?

seromenho:
@childish-sambino Thank you. With #1172 I can make my own wrapper instead. 
I saw #1128 was also requesting for something like this. I'll leave it to you to decide if this add some value or not.

packages/mail/package.json
childish-sambino: It's not an added dependency since it's already included with helpers. Explicitly adding it to the dependencies here is the best practice. I see no issue.

packages/mail/src/classes/mail-service.js
childish-sambino: Prefer to throw an exception if it's not an object instead of just returning.

packages/mail/src/classes/mail-service.js
```diff
@@ -184,8 +197,11 @@ class MailService {
         data.substitutionWrappers = this.substitutionWrappers;
       }
 
+      //Deep merge default data
+      const dataWithDefaults = deepmerge(this.defaultData, data);
```
childish-sambino, line 201: Tests needed to ensure the default data is actually getting merged.

packages/mail/src/classes/mail-service.js
```diff
@@ -110,6 +112,17 @@ class MailService {
     });
   }
 
+  /**
+   * Set default data to for all requests. Can be overwritten by incoming data on send
+   */
+  setDefaultData(data) {
```
childish-sambino, line 118: Example usage also needed.

packages/mail/src/classes/mail-service.js
seromenho: Need some help on testing, documentation is not very clear to me https://github.com/sendgrid/sendgrid-nodejs/blob/main/CONTRIBUTING.md#testing
Seems there is no prism:install and prism command. Is there a reason why this is not on devDependencies?
Thanks

childish-sambino: Testing section updated with latest instructions.

childish-sambino:
Closing due to inactivity. Please ping here if work is to continue.

Issue #950, opened 2019-07-03
Add keep-alive support


bcanseco:
#### Issue Summary

First off: this repo (and Sendgrid overall) is fantastic! ❤️ 

I'm using `@sendgrid/mail` and was interested in the behavior for handling [keep-alive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection#Syntax).
I noticed that the core `@sendgrid/client` package uses the [`request`](https://npmjs.com/package/request) library behind the scenes to hit the API.

https://github.com/sendgrid/sendgrid-nodejs/blob/68b7c45234558207c0a5263a6a45513810cb3afe/packages/client/src/classes/client.js#L6

https://github.com/sendgrid/sendgrid-nodejs/blob/68b7c45234558207c0a5263a6a45513810cb3afe/packages/client/src/classes/client.js#L115

There's a `forever` option that could be set to `true` in the `defaultRequest` configuration.

https://github.com/sendgrid/sendgrid-nodejs/blob/68b7c45234558207c0a5263a6a45513810cb3afe/packages/client/src/classes/client.js#L37-L43

This would enable `keepAlive` in the underlying http(s) agent.

#### Prior art

* [Stripe](http://stripe.com)'s Node.js library [uses keep-alive](https://github.com/stripe/stripe-node/search?q=keepalive&unscoped_q=keepalive) in the underlying http(s) agent by default
* Twilio does too, but requires opt-in [[1]](https://github.com/twilio/twilio-node/pull/315) [[2]](https://github.com/twilio/twilio-node/pull/386) which is probably what we would want to do here

#### Further reading

* https://lob.com/blog/use-http-keep-alive
* https://en.wikipedia.org/wiki/HTTP_persistent_connection
* https://en.wikipedia.org/wiki/Keepalive
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive

marcghorayeb:
I was looking for this as well. I looked a bit through the code and turns out, you can do this programmatically:

```
const httpsAgent = https.Agent({ maxSockets: 20, keepAlive: true, ...options });
SendGrid.client.setDefaultRequest('agent', httpsAgent);
```

A built-in solution would be better though.

childish-sambino:
Pull requests to add this feature are welcome and will be reviewed based on priority, but Twilio SendGrid is not actively building new functionality for the library.

thinkingserious:
[...]

PR #1226, opened 2020-12-04
feat: add keep alive support

khshah6:
Resolves: #950 

**Description**
This fix enables consumers to reuse connections by supporting [keep-alive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection#Syntax).

When `keepAlive` is set to true, SendGrid will use [forever-agent](https://github.com/request/forever-agent) (HTTP Agent that keeps socket connections alive between keep-alive requests).

This has been tested in a simple node request. End users do not have to provide any additional information when calling SendGrid for a client or to send mail and as such has no breaking change.

Linked issue includes examples of similar changes made in Stripe and Twilio apis.

### Checklist

If you have questions, please file a [support ticket](https://support.sendgrid.com), or create a GitHub Issue in this repository.

### DIFF
--- a/packages/mail/src/classes/mail-service.js
+++ b/packages/mail/src/classes/mail-service.js
@@ -40,6 +40,17 @@ class MailService {
     return this;
   }
 
+   /**
+   * Reuse connections
+   */
+  setKeepAlive(keepAlive) {
+    if (keepAlive === false) {
+      return;
+    }
+
+    this.client.setDefaultRequest('agent', new Agent({ keepAlive: true }));
+  }
+  
   /**
    * Twilio Email Auth passthrough for convenience.
    */
--- a/packages/mail/src/classes/mail-service.js
+++ b/packages/mail/src/classes/mail-service.js
@@ -41,7 +41,7 @@ class MailService {
     return this;
   }
 
-   /**
+  /**
    * Reuse connections
    */
   setKeepAlive(keepAlive) {
--- a/packages/mail/src/classes/mail-service.js
+++ b/packages/mail/src/classes/mail-service.js
@@ -3,6 +3,7 @@
 /**
  * Dependencies
  */
+const Agent = require('http').Agent;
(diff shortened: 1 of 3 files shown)


packages/mail/src/classes/mail-service.js
```diff
@@ -40,6 +41,17 @@ class MailService {
     return this;
   }
 
+  /**
+   * Reuse connections
+   */
+  setKeepAlive(keepAlive) {
+    if (keepAlive === false) {
+      return;
```
eshanholtz, line 49: What about turning off keep alive? If this just returns, you can never "unset" it.

packages/mail/src/classes/mail-service.js
```diff
@@ -40,6 +41,17 @@ class MailService {
...
+    }
+
+    this.client.setDefaultRequest('agent', new Agent({ keepAlive: true }));
```
eshanholtz, line 52: I think the key needs to be `httpsAgent`, according to the [documentation](https://github.com/axios/axios#request-config). This logic should be moved to the underlying `Client` with a separate helper method that we can call from here, similar to calling [`this.client.setApiKey`](https://github.com/khshah6/sendgrid-nodejs/blob/khshah6-add_keep_alive_support/packages/client/src/classes/client.js#L38) from [`setApiKey`](https://github.com/sendgrid/sendgrid-nodejs/pull/1226/files#diff-a4d6dd09858240682deb6dc3cf67bdda6a4b3b5e041537940d56a5f42b7c9a51R38).

khshah6: sure, good call out

packages/mail/src/classes/mail-service.js
khshah6: `keepAlive` is already false by default, its an opt-in feature

JenniferMah:
Closing until PR feedback is addressed.

Issue #986, opened 2019-09-18
Unable to Instantiate a new Client class


danhlee329:
#### Issue Summary

Currently I'm able to [manually create client instances](https://github.com/sendgrid/sendgrid-nodejs/tree/master/packages/client#instantiate-client-instances-manually) in javascript, but am not able to do so in Typescript.

#### Steps to Reproduce

```
import { Client } from '@sendgrid/client';
const clientInstance = new Client();
```
Error : Cannot use 'new' with an expression whose type lacks a call or construct signature.

#### Technical details:
- `@sendgrid/client` Version: 6.4.0
- `Node.js` Version: 10.16.3

dfcowell:
Also encountering this issue. Very frustrating for Typescript users.

manusis:
We also need support for creating new instance of SendGrid/Mail fo a new API key. In Nodejs modules are singleton objects, so setApiKey would not work as it would cause race around conditions when multiple API keys are used.

mgummelt:
I just ran into this problem as well.  Would be great to get a fix.  I'm pretty sure the problem is that this line: https://github.com/sendgrid/sendgrid-nodejs/blob/4cfbba40bf63142605d53fac7c8409894a8fbee1/packages/client/src/client.d.ts#L37

Should instead read:

`declare const client: Client & { Client: typeof Client }`

childish-sambino:
PR submitted to fix this.

PR #1227, opened 2020-12-05
chore: enhance client-related typings

kyle221b:
# Fixes #

The MailService has a method to set a client, but it isn't exposed in typings. Adding this so there is an easier way to customize underlying agents used (for keep-alive and other connection-pooling settings) without having compilation problems (or subverting the type system by just using 'any')

### Checklist
- [X] I have made a material change to the repo (functionality, testing, spelling, grammar)
- [X] I have read the [Contribution Guidelines](https://github.com/sendgrid/sendgrid-nodejs/blob/main/CONTRIBUTING.md) and my PR follows them
- [X] I have titled the PR appropriately
- [X] I have updated my branch with the main branch
- [X] I have added tests that prove my fix is effective or that my feature works

If you have questions, please file a [support ticket](https://support.sendgrid.com), or create a GitHub Issue in this repository.

### DIFF
--- a/packages/client/src/client.d.ts
+++ b/packages/client/src/client.d.ts
@@ -46,5 +46,8 @@ declare class Client {
   request(data: ClientRequest, cb?: (err: ResponseError, response: [ClientResponse, any]) => void): Promise<[ClientResponse, any]>;
 }
 
-declare const client: Client & { Client: typeof Client };
+declare const client: Client;
+// @ts-ignore
 export = client
+
+export {Client};
--- a/packages/helpers/classes/request.d.ts
+++ b/packages/helpers/classes/request.d.ts
@@ -1,3 +1,5 @@
+import * as https from 'https';
+
 type HttpMethod = 'get'|'GET'|'post'|'POST'|'put'|'PUT'|'patch'|'PATCH'|'delete'|'DELETE';
 
 export default interface RequestOptions<TData = any, TParams = object> {
@@ -7,4 +9,5 @@ export default interface RequestOptions<TData = any, TParams = object> {
     qs?: TParams;
     body?: TData;
     headers?: object;
+    httpsAgent?: https.Agent;
 }
--- a/packages/mail/src/mail.d.ts
+++ b/packages/mail/src/mail.d.ts
@@ -1,4 +1,4 @@
-import { Client } from "@sendgrid/client";
+import {Client} from "@sendgrid/client";
 import {ClientResponse} from "@sendgrid/client/src/response";
 import {ResponseError} from "@sendgrid/helpers/classes";
 import {MailDataRequired} from "@sendgrid/helpers/classes/mail";
--- a/packages/mail/src/mail.d.ts
+++ b/packages/mail/src/mail.d.ts
@@ -1,3 +1,4 @@
+import { Client } from "@sendgrid/client";
 import {ClientResponse} from "@sendgrid/client/src/response";
 import {ResponseError} from "@sendgrid/helpers/classes";
 import {MailDataRequired} from "@sendgrid/helpers/classes/mail";
@@ -7,6 +8,11 @@ declare class MailService {
    * SendGrid API key passthrough for convenience.
    */
   setApiKey(apiKey: string): void;
+
+  /**
+   * Client to use for invoking underlying API
+   */
+  setClient(client: Client): void;
 
   /**
    * Twilio Email Auth passthrough for convenience.
--- a/test/typescript/client.ts
+++ b/test/typescript/client.ts
@@ -1,4 +1,5 @@
 import Client = require("@sendgrid/client");
+import * as https from 'https';
 
 // Test setApiKey() method
 Client.setApiKey("MY_SENDGRID_API_KEY");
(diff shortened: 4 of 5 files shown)


packages/client/src/client.d.ts
```diff
@@ -46,5 +46,8 @@ declare class Client {
   request(data: ClientRequest, cb?: (err: ResponseError, response: [ClientResponse, any]) => void): Promise<[ClientResponse, any]>;
 }
 
-declare const client: Client & { Client: typeof Client };
+declare const client: Client;
+// @ts-ignore
```
kyle221b, line 50: This matches style of mail.d.ts file. Part that I wanted to change here mainly was the `{ Client: typeof Client }` piece. I get the feeling that the goal there was to expose the class as a type, but we do `typeof` on the class itself, instead of an instance of it, so it ends up being a type with just a `prototype` field. Let me know if I'm missing something here.

kyle221b:
@thinkingserious any chance you or someone from your team could take a look at this?

packages/client/src/client.d.ts
thinkingserious: I believe this change might break folks: https://github.com/sendgrid/sendgrid-nodejs/issues/986

packages/client/src/client.d.ts
```diff
@@ -46,5 +46,8 @@ declare class Client {
...
 export = client
+
+export {Client};
```
thinkingserious, line 53: We would need to revert this change also if we move back to `declare const client: Client & { Client: typeof Client };`

packages/helpers/classes/request.d.ts
```diff
@@ -7,4 +9,5 @@ export default interface RequestOptions<TData = any, TParams = object> {
     qs?: TParams;
     body?: TData;
     headers?: object;
+    httpsAgent?: https.Agent;
```
thinkingserious, line 12: I believe this data is stored in `headers`.

thinkingserious (commented):
Thank you for the PR @kyle221b!

packages/client/src/client.d.ts
kyle221b: yeah the goal here was to export `Client` as not just a value with type `typeof Client`, but to export the class itself, which could be used as a value or a type. im pretty sure this will only enhance the existing functionality. currently, users could do:
`const a = new Client()`

but the only way they could pass the actual type around is by doing a default import of the singleton client and then doing a `typeof` on it.

am i missing something here?

packages/client/src/client.d.ts
kyle221b: left a comment above about why I did this. let me know what you think

packages/helpers/classes/request.d.ts
kyle221b: this is a top-level field on an `axios` request. i don't think we'd want to pass it in headers b/c it won't actually get sent across the wire, it's just connection settings essentially.

if we were to add it to headers, then there would need to be a code change in the JS file to filter it out and pass it as a top-level field in the axios request

packages/client/src/client.d.ts
thinkingserious: I don't quite understand why this method is better than what we have. Perhaps a good way to illustrate is to add a sample quick start [here](https://github.com/sendgrid/sendgrid-nodejs/tree/main/docs/use-cases).

kyle221b: To give a quick code example:

```
import { Client } from '@sendgrid/client';

// This is possible in both versions
const client = new Client();

// This (I think) is only possible in the new version since Client is only a value in the old version and not a type
function somethingThatTakesInAClient(client: Client) {
...
}
```

kyle221b: while it isn't necessary, you do it for most of your other types that I've seen, and if you are going to export functionality, it seems to me it would also make sense to export the type

kyle221b: to further illustrate, in the current version, we could accomplish the same by doing:
```
// Note we are importing the singleton client exported by the lib as well as the class itself (as a value)
import client, { Client } from '@sendgrid/client';

// We can do a typeof on the singleton instance. This is equivalent to what is in the last example. For this one you need // to import a singleton instance of the client just to get the type
function somethingThatTakesInAClient(arg: typeof client) {
...
}

// This compiles, but the behavior isn't what you expect. Since we are doing typeof on the class, the only members are // things on the function prototype, so we can't access methods like 'setApiKey' for example
function anotherThingThatLooksLikeItTakesInAClientButDoesntReally(arg: typeof Client) {
...
}
```

kyle221b: Here's an example where you use the proposed pattern: https://github.com/sendgrid/sendgrid-nodejs/blob/main/packages/mail/src/mail.d.ts#L39-L42.

thinkingserious:
@kyle221b,

Your reasoning on [this thread](https://github.com/sendgrid/sendgrid-nodejs/pull/1227#pullrequestreview-545387327) makes sense to me. However, I'm still concerned about the points made in [this PR](https://github.com/sendgrid/sendgrid-nodejs/issues/986).

@danhlee329, @dfcowell, @manusis, @mgummelt -- do you have any thoughts on this issue? Thanks!

With best regards,

Elmer

kyle221b:
Anybody have thoughts on this?

JenniferMah (approved):

PR #1257, opened 2021-02-24
fix: Node 6 build failure

thinkingserious:
This change ensures that we don't pull in [Chai v4.3.0](https://www.chaijs.com/releases/) (which no longer supports Node 6) with `npm install` or `yarn install`.

### Checklist


JenniferMah (approved):
LGTM

PR #1271, opened 2021-04-29
chore: add signed webhook test for multi event payload

### LINKED ISSUE #1238: Event Webhook only works for a single event payload
### Issue Summary
The helper library to validate the webhook signature only works when the payload has a single event. when the payload is an array of multiple JSON objects the signature always returns false. as seen here: https://github.com/sendgrid/sendgrid-nodejs/blob/main/packages/eventwebhook/src/eventwebhook.spec.js there is only 1 test and this tests uses a payload with a single event. I've opened a ticket with the sendgrid support team who told me to open an issue here stating the problem.

### Steps to Reproduce
1. setup a signed webhook as seen here: https://sendgrid.com/docs/for-developers/tracking-events/getting-started-event-webhook-security-features/
2. if sendgrid sends a single event per request everything is fine and the verifySignature function always returns true
3. if sendgrid aggregates 2 or more events in a single request the verifySignature function returns false

### Code Snippet
Basically the same as here: https://github.com/sendgrid/sendgrid-nodejs/blob/main/packages/eventwebhook/src/eventwebhook.spec.js with the payload being a 2 object array.

### Exception/Log
No exception, just returns false

### Technical details:
I have an example payload, signature and timestamp received that I shared with the support team when creating the ticket but since the payload contains sensitive information I'm not pasting that here. I'll be glad to do so via a private chat or something similar. 
To quote the answer that I got from the technical support: "I have looked into this further and would agree this could be a library specific issue with the configuration since a single event works. I also reviewed the payload and do not see anything that is wrong. "

I'm running this code in a aws lambda running Node.js 12.x. To be able to use the webhook helper lib I copied this file into my lambda env https://github.com/sendgrid/sendgrid-nodejs/blob/main/packages/eventwebhook/src/eventwebhook.js and used it as shown in its respective tests (as linked above)

shwetha-manvinkurke:
Add example to show how to process payloads with multiple events for signed webhook verification. 

Note: The webhook verification logic itself works fine with the raw request body in the example shown [here](https://github.com/sendgrid/sendgrid-nodejs/blob/main/docs/use-cases/event-webhook.md). 

Fixes #1238 

### Checklist

### DIFF
--- a/packages/eventwebhook/src/eventwebhook.spec.js
+++ b/packages/eventwebhook/src/eventwebhook.spec.js
@@ -18,13 +18,52 @@ describe('EventWebhook', () => {
     ]
   ) + '\r\n'; // Be sure to include the trailing carriage return and newline!
 
+  const PUBLIC_KEY_MULTIPLE = 'MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEQ4LIFtWztlsF7skFqOncjD1lun4H5w8XOhyOArHW9RcIx/FfEzx6cikC/yPfUvwaX/JScE7Fc9CJD2afQ9Ok3Q==';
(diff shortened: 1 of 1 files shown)


childish-sambino (approved):
Should link this to the open issue.

PR #1320, opened 2021-12-01
chore: adjust 'packages/mail/src/mail.d.ts'

collierrgbsitisfise:
Error for callBack function of 'send' and 'sendMultiple' methods is not
described as nullable, but it should cause error is null when emails
where successfully delivered

### DIFF
--- a/packages/mail/src/mail.d.ts
+++ b/packages/mail/src/mail.d.ts
@@ -32,12 +32,12 @@ declare class MailService {
   /**
    * Send email
    */
-  send(data: MailDataRequired | MailDataRequired[], isMultiple?: boolean, cb?: (err: Error | ResponseError, result: [ClientResponse, {}]) => void): Promise<[ClientResponse, {}]>;
+  send(data: MailDataRequired | MailDataRequired[], isMultiple?: boolean, cb?: (err: Error | ResponseError | null, result: [ClientResponse, {}]) => void): Promise<[ClientResponse, {}]>;
 
   /**
    * Send multiple emails (shortcut)
    */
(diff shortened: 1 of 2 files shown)


packages/mail/src/mail.d.ts
```diff
@@ -32,12 +32,12 @@ declare class MailService {
...
    * Send email
    */
-  send(data: MailDataRequired | MailDataRequired[], isMultiple?: boolean, cb?: (err: Error | ResponseError, result: [ClientResponse, {}]) => void): Promise<[ClientResponse, {}]>;
+  send(data: MailDataRequired | MailDataRequired[], isMultiple?: boolean, cb?: (err: Error | ResponseError | null, result: [ClientResponse, {}]) => void): Promise<[ClientResponse, {}]>;
 
   /**
    * Send multiple emails (shortcut)
    */
-  sendMultiple(data: MailDataRequired, cb?: (error: Error | ResponseError, result: [ClientResponse, {}]) => void): Promise<[ClientResponse, {}]>;
+  sendMultiple(data: MailDataRequired, cb?: (error: Error | ResponseError | null, result: [ClientResponse, {}]) => void): Promise<[ClientResponse, {}]>;
```
JenniferMah, line 40: Hi @collierrgbsitisfise! Thanks for the contribution. Can you add some additional tests to verify that error is null when emails are successfully delivered?

JenniferMah (dismissed):

packages/mail/src/mail.d.ts
collierrgbsitisfise: @JenniferMah done, add more test cases for mail functionality

JenniferMah (approved):
Thanks for adding tests @collierrgbsitisfise!

PR #1329, opened 2022-01-28
chore: migrate to gh actions

shwetha-manvinkurke:


JenniferMah (commented):
Could you also remove the .codeclimate.yml file since we don't run code climate anymore?

eshanholtz (approved):

.travis.yml
```diff
@@ -1,34 +0,0 @@
-language: node_js
-node_js:
-- 10
-env:
-- version=6
-- version=8
-- version=10
-- version=lts
-- version=latest
```
childish-sambino, line 9: All of the SendGrid helpers that use `make test-docker` don't rely on the language version that's installed. They rely on the `version` env var which is used in their Dockerfile. The GH action needs to set the `version` env var since right now it will always just use `lts`:
https://github.com/sendgrid/sendgrid-nodejs/blob/32516106d27e8e228a09e147bd6d8150f6c6f4a1/Dockerfile#L1-L2

PR #1330, opened 2022-01-31
test: set the right version for tests

shwetha-manvinkurke:
https://github.com/sendgrid/sendgrid-nodejs/pull/1329#discussion_r796108313

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -30,16 +30,10 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_AUTH_TOKEN }}
 
-      - name: Set up Node
-        uses: actions/setup-node@v2
-        with:
-          node-version: ${{ matrix.node }}
-
       - name: Reset env vars
         if: env.version == 'lts/*'
         run: echo "version=lts" >> "$GITHUB_ENV"
 
-      - run: npm install
       - name: Run Tests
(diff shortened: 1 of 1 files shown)


.github/workflows/test-and-deploy.yml
```diff
@@ -33,6 +35,10 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
 
+      - name: Reset env vars
+        if: env.version == 'lts/*'
+        run: echo "version=lts" >> "$GITHUB_ENV"
+
       - run: npm install
```
childish-sambino, line 42: Don't need this step since it's run inside docker.

.github/workflows/test-and-deploy.yml
```diff
@@ -33,6 +35,10 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
```
childish-sambino, line 36: Don't need this step since it's run inside docker.

.github/workflows/test-and-deploy.yml
```diff
@@ -33,6 +35,10 @@ jobs:
...
 
+      - name: Reset env vars
+        if: env.version == 'lts/*'
```
childish-sambino, line 39: Can just change the matrix to use `lts` instead of `lts/*`.

shwetha-manvinkurke: That doesn't work

childish-sambino (approved):

PR #1336, opened 2022-02-09
chore: don't test for EOL node versions

camelmasa:
I think we don't need to test EOL node versions. How is this?


childish-sambino:
Given the library still works with these older versions, we'll continue to support their usage until we no longer can. Not recommending anyone stay on these old versions, but we'll continue to test against them.

PR #1351, opened 2022-03-16
fix: Fix typings for eventwebhook.d.ts PublicKey

Cellule:
# Fixes

Installing `@sendgrid/eventwebhook` in a project using typescript cannot properly compile because `starkbank-ecdsa` does not provide any typings and none are available in DefinitelyTyped project either.
Right now the only way to use the module is to ourselves provide a definition of `starkbank-ecdsa` OR patch-package `@sendgrid/eventwebhook`

I've decided to change the import statement with a minimal interface of `PublicKey` which should be enough to use the package properly in a typescript project.

Technically speaking, `@types/node` is also required because you use `Buffer` in the types, but a typescript project without it would be weird so I think we can let that one as-is.

### Checklist


childish-sambino (approved):
LGTM

PR #1355, opened 2022-03-30
[TEST DON'T MERGE] Updating dependencies and resolving build failures

rakatyal:


packages/inbound-mail-parser/package.json
```diff
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "@sendgrid/helpers": "^7.6.2",
-    "mailparser": "^2.3.4"
+    "mailparser": "^3.3.0"
```
JenniferMah, line 30: Is there a reason we can't upgrade to [v3.4.0](https://github.com/nodemailer/mailparser/releases/tag/v3.4.0)?

packages/subscription-widget/package.json
```diff
@@ -13,7 +13,7 @@
     "body-parser": "^1.15.2",
     "express": "^4.14.0",
     "morgan": "^1.9.1",
-    "nodemon": "^1.9.2",
+    "nodemon": "^2.0.3",
```
JenniferMah, line 16: Is there a reason we can't upgrade to [v2.0.15](https://github.com/remy/nodemon/releases/tag/v2.0.15)?

packages/inbound-mail-parser/package.json
rakatyal: Sorry yeah, this is still a draft PR. I was just testing to fix the build failures. I will update the packages to the latest :)

rakatyal: Done.

packages/subscription-widget/package.json
rakatyal: Done.

lerna.json
```diff
@@ -1,5 +1,6 @@
 {
   "npmClient": "yarn",
+  "npmClientArgs": ["--ignore-engines"],
```
childish-sambino, line 3: Don't know how I feel about this. Should we just drop support for Node 6/8 instead?

rakatyal: I was gonna bring this up for discussion tomorrow but I am fine either way. What's the impact of dropping Node 6/8? If we go that route, should we drop support for all the packages in this repo or just the affected ones in this PR?

rakatyal: Looks like npm considers the ["engines" flag](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines) as advisory and will only generate warnings in this scenario, but yarn gives out errors. I would support dropping Node 6/8 if we feel this change is a risk which our tests cannot catch when testing with Node 6/8. Or either there is no significant impact on dropping support for Node 6/8. Thoughts?

childish-sambino: We've seen issues in the past with other Node libs where we had to use older implementations of core functionality because we supported older Node versions. Leads to maintenance and TODOs. Prefer to drop Node 6/8 everywhere as this meets the criteria where we're doing workarounds to support older versions. Will check with product.

PR #1371, opened 2022-08-15
Issue: types are wrong when calling send() with an array

jameshfisher:
This is a bit of a guess at the "correct" type.


jacksonloper:
.... is this still a thing?

PR #1373, opened 2022-09-05
Issue: what is "customArgs"? Does it correspond to "unique_args" in the API?

jameshfisher:
I want to set [unique arguments](https://docs.sendgrid.com/for-developers/sending-email/unique-arguments) using this library. However, I can't see any way to do that. I can see a field called `customArgs`, but it's unclear whether this is the same thing. Can you point me to documentation on field?


jameshfisher:
(That said, "unique args" is a bizarre name, and should be called "tags" or "private metadata". But the most important thing is consistency - just call it one thing!)

PR #1387, opened 2023-11-06
fix(axios): updated axios version to fix critical vulnerability

joserodriguezjll:
# Fixes #

- Updated axios version to fix critical vulnerability

### Checklist
- [x] I acknowledge that all my contributions will be made under the project's license
- [x] I have made a material change to the repo (functionality, testing, spelling, grammar)
- [x] I have read the [Contribution Guidelines](https://github.com/sendgrid/sendgrid-nodejs/blob/main/CONTRIBUTING.md) and my PR follows them
- [x] I have titled the PR appropriately
- [x] I have updated my branch with the main branch
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added the necessary documentation about the functionality in the appropriate .md file
- [x] I have added inline documentation to the code I modified

If you have questions, please file a [support ticket](https://support.sendgrid.com).

PR #1390, opened 2023-11-20
feat: Add data residency for eu and global regions

shrutiburman:
# Fixes [DII-1230](https://issues.corp.twilio.com/browse/DII-1230)

This PR:
- Adds a setter `setDataResidency()` which allows the users to switch to EU sub region, reidrecting their requests to `api.eu.sendgrid.com`
- Upgrades the testing pipeline to node v10+ ([eslint](https://github.com/sendgrid/sendgrid-nodejs/blob/b8125d86134157a462c2e24816cfa56677ab76a8/package.json#L29C109-L29C147) and typescript do not work well on <10 versions)
- Upgrades typescript, which was causing the pipleine failure for very long

### Checklist
- [x] I have added the necessary documentation about the functionality in the appropriate .md file

If you have questions, please file a [support ticket](https://support.sendgrid.com).

### DIFF
--- a/docs/use-cases/data-residency-set-hostname.md
+++ b/docs/use-cases/data-residency-set-hostname.md
@@ -1,8 +1,8 @@
-# Choosing a hostname to send messages to
+# Choosing a data-residency to send messages to
 
 Use the `setDataResidency` setter to specify which host to send to:
 
-Send to EU (hostname: `https://api.eu.sendgrid.com/`)
+Send to EU (data-residency: `https://api.eu.sendgrid.com/`)
 ```js
 const sgMail = require('@sendgrid/mail');
 sgMail.setDataResidency('eu');
@@ -16,7 +16,7 @@ const msg = {
 sgMail.send(msg);
 ```
 Send to Global region, this is also the default host, if the setter is not used
-(hostname: `https://api.sendgrid.com/`)
+(data-residency: `https://api.sendgrid.com/`)
 ```js
 const sgMail = require('@sendgrid/mail');
 sgMail.setDataResidency('global');
@@ -32,7 +32,6 @@ sgMail.send(msg);
 
 ## Limitations
 
-1. Setting the API Key (via `client.setApiKey()`) or Twilio Authentication (via `client.setTwilioEmailAuth()`) will override the hostname to default value. Use the setter call after this set-up.
-2. Emails can only be sent to two hosts for now; 'eu' (https://api.eu.sendgrid.com/) and 'global' (https://api.eu.sendgrid.com/)
-2. The default hostname is https://api.sendgrid.com/
+1. Emails can only be sent to two hosts for now; 'eu' (https://api.eu.sendgrid.com/) and 'global' (https://api.eu.sendgrid.com/)
+2. The default data-residency is https://api.sendgrid.com/
 3. The valid values for `region` in `client.setDataResidency(region)` are only `eu` and `global`. Case-sensitive.
--- a/docs/use-cases/data-residency-set-hostname.md
+++ b/docs/use-cases/data-residency-set-hostname.md
@@ -32,6 +32,7 @@ sgMail.send(msg);
 
 ## Limitations
 
-1. Emails can only be sent to two hosts for now; 'eu' (https://api.eu.sendgrid.com/) and 'global' (https://api.eu.sendgrid.com/)
+1. Setting the API Key (via `client.setApiKey()`) or Twilio Authentication (via `client.setTwilioEmailAuth()`) will override the hostname to default value. Use the setter call after this set-up.
+2. Emails can only be sent to two hosts for now; 'eu' (https://api.eu.sendgrid.com/) and 'global' (https://api.eu.sendgrid.com/)
 2. The default hostname is https://api.sendgrid.com/
 3. The valid values for `region` in `client.setDataResidency(region)` are only `eu` and `global`. Case-sensitive.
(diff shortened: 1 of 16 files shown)


manisha1997 (approved):
LGTM

packages/client/src/client.spec.js
```diff
@@ -3091,3 +3092,46 @@ describe('test_whitelabel_links__link_id__subuser_post', () => {
...
+    testClient.setDataResidency('');
+    expect(consoleWarnSpy.calledOnce).to.equal(true);
+  });
+  it('should give a warning if the provided value is null', () => {
+    testClient.setDataResidency(null);
+    expect(consoleWarnSpy.calledOnce).to.equal(true);
+  });
+  it('should give precedence to the order of execution', () => {
+    testClient.setDataResidency('eu');
+    testClient.setApiKey('SG.1234567890');
```
gladysmae08, line 3130: why would `setApiKey` override the `hostname`? If it really needs to, is this clearly documented somewhere? But I personally think this should not change the `hostname`.

gladysmae08 (commented):
CR

packages/client/src/client.spec.js
shrutiburman: For some reason, setApiKey() also [sets the baseUrl](https://github.com/sendgrid/sendgrid-nodejs/blob/b8125d86134157a462c2e24816cfa56677ab76a8/packages/client/src/classes/client.js#L41) while in action. I do not know why exactly was this done, but it looks like for each auth call for [Sendgrid](https://github.com/sendgrid/sendgrid-nodejs/blob/b8125d86134157a462c2e24816cfa56677ab76a8/packages/client/src/classes/client.js#L39) or [Twilio](https://github.com/sendgrid/sendgrid-nodejs/blob/b8125d86134157a462c2e24816cfa56677ab76a8/packages/client/src/classes/client.js#L48), it was needed.

shrutiburman: I can add this in the **Limitations** section in usage doc.

shrutiburman: Done ^

gladysmae08: I see. I think that was done to automatically switch between the twilio vs sendgrid endpoint based on the type of auth used. It makes sense for that specific use case but I think we should try to elegantly handle this on the regional sendgrid side. How does it work for other libraries?

I'm thinking something along the lines of:
- set to global endpoint if `baseUrl` is not yet set to sendgrid
- leave the value as is if it's already using one of the sendgrid endpoints (eu or global).

OR we can track the region that was set and `setApiKey` will set the `baseUrl` using the correct region.

cc: @tbischel in case you have any input

tiwarishubham635 (approved):
LGTM!

packages/client/src/client.spec.js
shrutiburman: Other libraries don't have this situation, they have different classes for Twilio and Sendgrid, inheriting the same parent class.

tbischel: I agree with Gladys, this becomes a bad user experience as is when any combination of setApiKey and setting the data residency happen in different orders.  I'd be inclined to track the region, where setApiKey uses the correct region, and the default region is global.

tbischel (changes requested):
I think we need to resolve this usability gap on the setApiKey and setDataResidency before we release.

docs/use-cases/data-residency-set-hostname.md
```diff
@@ -0,0 +1,38 @@
...
+  subject: 'Hello world',
+  text: 'Hello plain world!',
+  html: '<p>Hello HTML world!</p>',
+};
+sgMail.send(msg);
+```
+
+## Limitations
+
+1. Setting the API Key (via `client.setApiKey()`) or Twilio Authentication (via `client.setTwilioEmailAuth()`) will override the hostname to default value. Use the setter call after this set-up.
```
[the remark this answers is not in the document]
Be3751: @tbischel @gladysmae08 
Hi, I hope my [PR](https://github.com/sendgrid/sendgrid-nodejs/pull/1386) offers a solution to this issue. Please take a look and I would appreciate any feedback.

gladysmae08 (commented):
CR

mvanzoest:
Hi, would it be possible to open a separate PR for the version upgrades? A number of people are awaiting #1387 in order to complete their security upgrades.

shrutiburman:
> I think we need to resolve this usability gap on the setApiKey and setDataResidency before we release.

We have resolved the discrepancy in the latest commits. The limitation no longer exists.

packages/client/src/classes/client.js
```diff
@@ -38,8 +44,10 @@ class Client {
 
   setApiKey(apiKey) {
     this.auth = 'Bearer ' + apiKey;
-    this.setDefaultRequest('baseUrl', SENDGRID_BASE_URL);
-
+    // this means that region was never set before
+    if (this.sendgrid_region == '') {
+      this.setDefaultRequest('baseUrl', SENDGRID_BASE_URL);
+    }
```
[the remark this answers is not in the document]
shrutiburman: Yup, I have set the default region as 'global', we can reset based on that now. Thanks.

docs/use-cases/data-residency-set-hostname.md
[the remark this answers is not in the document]
tbischel: this won't introduce breaking changes for existing customers, right?

gladysmae08 (approved):
LGTM

docs/use-cases/data-residency-set-hostname.md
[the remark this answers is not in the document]
shrutiburman: This wont introduce a breaking change, as the `region` attribute that's newly introduced has a default value, and will only get updated if the setter for data residency is called.

tbischel (approved):

PR #1428, opened 2025-02-18
fix: Support 204 response in axios

lopenchi:
# Fixes #

Related PR: https://github.com/sendgrid/sendgrid-nodejs/pull/1327

fix: Support 204 response in axios 

### Checklist

If you have questions, please file a [support ticket](https://support.sendgrid.com).

### DIFF
--- a/packages/client/src/classes/client.js
+++ b/packages/client/src/classes/client.js
@@ -154,7 +154,11 @@ class Client {
     data = this.createRequest(data);
 
     const promise = new Promise((resolve, reject) => {
-      axios(data)
+      axios(data, {
+        validateStatus: function (status) {
+          return status >= 200 && status < 300 || status === 204;
+        },
+      })
         .then(response => {
           return resolve([
             new Response(response.status, response.data, response.headers),
--- a/packages/client/src/classes/client.js
+++ b/packages/client/src/classes/client.js
@@ -154,11 +154,7 @@ class Client {
     data = this.createRequest(data);
 
     const promise = new Promise((resolve, reject) => {
-      axios(data, {
-        validateStatus: function (status) {
-          return status >= 200 && status < 300 || status === 204;
-        },
-      })
+      axios(data)
         .then(response => {
           return resolve([
             new Response(response.status, response.data, response.headers),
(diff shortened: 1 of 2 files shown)


packages/client/src/classes/client.js
```diff
@@ -154,7 +154,11 @@ class Client {
     data = this.createRequest(data);
 
     const promise = new Promise((resolve, reject) => {
-      axios(data)
+      axios(data, {
+        validateStatus: function (status) {
+          return status >= 200 && status < 300 || status === 204;
```
bnb, line 159: I'm pretty sure `|| status === 204` here is not ever going to be hit because if `status` is `204`, the first condition will be met 👀

lopenchi: Yep but it is not working, you can see tests were skipped before. And it is how axios worked

bnb: I'm not sure I understand - if you _don't_ include `|| status === 204` in this line, the tests fail?

lopenchi: yes, axios do not support 204 responses if this line is not added

lopenchi: check this related PR: https://github.com/sendgrid/sendgrid-nodejs/pull/1327 where tests that returns 204 were skipped

bnb: Right, I understand. I'm not talking about the whole line, I'm specifically speaking about the contents of the line.

In code, here's what I mean:

```js
const status = 204

console.log(status >= 200 && status < 300 || status === 204) // returns true
```

```js
const status = 204

console.log(status >= 200 && status < 300) // returns true
```

both of these return true because _if_ `status` is `204`, status satisfies `status >= 200 && status < 300`, meaning that `|| status === 204` is a dead code path that will never be hit.

lopenchi: Got it! So I think this change is not needed.

What about not skipping the tests ? I think it has been already fixed in axios or prism

bnb: I think the change might still be needed - not sure either way! - but this line specifically should be tweaked so we're not shipping a dead code path that will never be executed. The solution to my original comment would be to simply drop `|| status === 204` from the line.

